PR #93, opened 2014-08-29
Introduce FormElement typified element

### LINKED ISSUE #92: Introduce FormElement typified element
Plan:

1. add new `FormElement` typified element class
2. subclass the following typified elements from it:
- `LabeledElement` (it covers both Radio Buttons and Checkboxes)
- `FileInput`
- `Select`
- `SelectOption`
- `TextInput`
3. add `setForm` method that will allow `Form` element to inject itself into each element it creates

aik099:
Closes #92


aik099:
Closing, because isn't needed by anyone currently.

PR #95, opened 2014-08-29
[WIP] Cache elements found by the locator via 'cacheLookup' annotation

### LINKED ISSUE #94: Cache elements found by the locator via 'cacheLookup' annotation
In the Java version of Selenium there was a `@cacheLookup` annotation that allowed to cache found `WebElement` class objects right inside of the `ElementLocator` class that searched for them.

If ElementLocator is to be used as a standalone component, then it might make sense, but the `PageFactory` is creating a separate locator class per each Page/ElementContainer class property it decorates and only accesses it once when element is being used, so caching doesn't really make sense.

We might implement the caching if somebody will be using ElementLocator in custom factory to save on calls made to the browser for element location. This cache however might result in stale element returning, that is no longer present in DOM.

aik099:
Closes #94

### DIFF
--- a/library/QATools/QATools/PageObject/Annotation/FindByAnnotation.php
+++ b/library/QATools/QATools/PageObject/Annotation/FindByAnnotation.php
@@ -98,6 +98,13 @@ class FindByAnnotation extends Annotation
 	public $using;
 
 	/**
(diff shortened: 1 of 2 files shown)


aik099:
Closing, because proxy (the only user of locator) is never asking it to find elements more than once.

PR #96, opened 2014-08-29
Documentation hosted on ReadTheDocs

aik099:
This branch contains documentation to be hosted on ReadTheDocs.Org website.


README.md
```diff
@@ -2,6 +2,7 @@
 
 [![Build Status](https://travis-ci.org/qa-tools/qa-tools.svg?branch=master)](https://travis-ci.org/qa-tools/qa-tools)
 [![HHVM Status](http://hhvm.h4cc.de/badge/qa-tools/qa-tools.svg)](http://hhvm.h4cc.de/package/qa-tools/qa-tools)
+[![Documentation Status](https://readthedocs.org/projects/qa-tools/badge/?version=latest)](https://readthedocs.org/projects/qa-tools/?badge=latest)
```
stof: the link when clicking on the doc badge should probably go to the rendered documentation rather than the RTD config

aik099: Nice catch, I'll fix that. I've just used Markdown code provided by the RTD themselves.

stof: I opened an issue on RTD to suggest this change too: https://github.com/rtfd/readthedocs.org/issues/1115

.travis.yml
```diff
@@ -25,9 +25,13 @@ before_script:
     - sudo sed -i -e "s,/var/www,$(pwd),g" /etc/apache2/sites-available/default
     - sudo /etc/init.d/apache2 restart
 
+    - sudo easy_install -U Sphinx
+    - sudo pip install sphinx_rtd_theme
+
 script:
     - mkdir -p build/logs
     - ./vendor/bin/paratest --coverage-clover build/logs/clover.xml
+    - sphinx-build -nW -b html -d docs/build/doctrees docs docs/build/html
```
stof: given that this does not depend on the PHP version, you might want to do it only for a single PHP version (or maybe even for a separate job building only the doc)

.travis.yml
```diff
@@ -25,9 +25,13 @@ before_script:
     - sudo sed -i -e "s,/var/www,$(pwd),g" /etc/apache2/sites-available/default
     - sudo /etc/init.d/apache2 restart
 
+    - sudo easy_install -U Sphinx
+    - sudo pip install sphinx_rtd_theme
+
 script:
     - mkdir -p build/logs
```
stof: creating this folder should go in before_script actually. It is not a testing step

.travis.yml
aik099: Since I have docs and code in same repo I need to build both of them at same time because I have no way of knowing what path was changed in the PR.

.travis.yml
aik099: Can you please make the corresponding PR, but targeting `develop` branch instead?

.travis.yml
stof: @aik099 I'm talking about different jobs in the matrix:

- code on PHP 5.3
- code on PHP 5.4
- ...
- code on HHVM
- Sphinx doc

Currently, you are building the Sphinx doc for each PHP version being tested, meaning you validate 5 times.

thus, having a separate job for the doc build would make it easier to see whether the test failure is a code failure on some PHP version or a syntax error in the doc

.travis.yml
stof: https://github.com/qa-tools/qa-tools/pull/113

.travis.yml
aik099: Now I get it. Yes, that would make sense. How to I create non-PHP dependant matrix element on PHP build?

stof: I have an idea for this, but I need to try it. I will send you a PR this evening

stof:
@aik099 what is the state of this PR ?

aik099:
This PR is for commenting purposes only until documentation is complete. I don't think that it's a good idea to merge it currently knowing that there are almost none useful docs in it. There are few pages written, but writing documentation from scratch is huge work, that needs a lot of time.

Maybe we should merge it now and then send subsequential PR for individual documentation chapters?

aik099:
What would be best way to solve current merge conflicts: rebase on top of `develop` or merge `develop`?

aik099:
The `rebase` gives me 4 conflicts on `.travis.yml`, but `merge` have no issues and merges all `develop` branch in.

stof:
@aik099 it would be a better idea to do individual PRs indeed. It is much easier than doing a single huge rewrite

Using a rebase would screw up the merges of other PRs, so I think merging is better.
However, there is no conflicts anymore (they were in the Travis config, and my latest change means that the travis config diff does not conflict anymore).

stof:
so you don't need to do anything for conflict resolution

aik099:
I've pushed merge result and will merge this PR after build stuff finishes.

aik099:
Merged.

PR #98, opened 2014-08-29
Remove rewind

### LINKED ISSUE #80: Remove usages of "rewind" method
After the `Iterator` interface in collections was changed to `ArrayObject` class (see #73) the need of doing `->rewind()` call after internal collection iteration (e.g. mass container injection or mass name setting in proxies) is gone. This is because the `getIterator` method we're using now is returning new iterator instance each time and rewinding previously used iterator won't affect next iterator, that will be used.

I propose to:

1. remove `->rewind()` calls
2. add test to verify that after internal iterations on collection that have several elements the interator index is 0

evangelion1204:
Old PR #91.

Closes #80

### DIFF
--- a/tests/QATools/QATools/HtmlElements/Proxy/TypifiedElementCollectionProxyTest.php
+++ b/tests/QATools/QATools/HtmlElements/Proxy/TypifiedElementCollectionProxyTest.php
@@ -54,6 +54,17 @@ class TypifiedElementCollectionProxyTest extends TypifiedElementProxyTest
 		$this->assertEquals(1, $this->element->proxyMe());
 	}
 
+	public function testInternalPointerPointingOnFirstElement()
+	{
+		$node_elements = $this->expectLocatorCallReturningTwoNodeElements();
+
+		foreach ( $node_elements as $node_element ) {
+			$this->driver->shouldReceive('getTagName')->with($node_element->getXpath())->andReturn('textarea');
+		}
+
+		$this->assertEquals(1, $this->element->getObject()->proxyMe());
+	}
+
 	/**
 	 * Creates a proxy.
 	 *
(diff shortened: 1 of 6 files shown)


tests/QATools/QATools/HtmlElements/Proxy/TypifiedElementCollectionProxyTest.php
```diff
@@ -47,6 +47,13 @@ public function testMethodForwardingSuccess()
 		$this->assertEquals(1, $this->element->proxyMe());
 	}
 
+	public function testInternalPointerPointingOnFirstElement()
+	{
+		$this->expectLocatorCallReturningTwoNodeElements();
+
+		$this->assertEquals(1, $this->element->getObject()->proxyMe());
```
evangelion1204: The `TypifiedElementCollectionProxyTest` is just a proxy to a collection and not to the elements inside. The test should just ensure that we point to the collection and not to the elements inside.

We could set `getObject` method to protected, then nobody could use it from outside. But this might be a new issue.

evangelion1204:
Rebased, see my comment above regarding the tests and `getObject`.

evangelion1204:
Can we merge this or is anything missing? I am currently not sure.

aik099:
I've totally forgot what this one is doing myself. Too much tests added to test single removed line. Will look into it shortly.

aik099:
Please squash and rebase, because after the assertion of element tag names/attributes was added to typified elements tests added here fails with `No Matching Expectation` found.

aik099:
Otherwise changes looks good. Suddenly it all became very clear to me, when I've viewed the changes in PhpStorm with my favorite Solarized color scheme and `hub checkout https://github.com/qa-tools/qa-tools/pull/98` command.

evangelion1204:
Squashed and rebased on develop again, currently waiting for coveralls :smile:

aik099:
Merged. Thanks, @evangelion1204 .

PR #99, opened 2014-08-29
Enhance getPage in PageFactory

### LINKED ISSUE #12: Add support for instantiating page classes from PageFactory class
Based on https://github.com/sensiolabs/BehatPageObjectExtension create factory to map page names to page class names (would allow substituting pages in complex tests).

Side effect - no IDE auto-complete unless inline ```@var``` annotation is used.

evangelion1204:
New PR to develop instead of master see #89.

Closes #12

### DIFF
--- a/library/QATools/QATools/PageObject/Config/Config.php
+++ b/library/QATools/QATools/PageObject/Config/Config.php
@@ -28,6 +28,7 @@ class Config implements IConfig
 	 */
 	protected $options = array(
 		'base_url' => '',
+		'page_namespace_prefix' => array('\\'),
 	);
 
 	/**
--- a/library/QATools/QATools/PageObject/PageFactory.php
+++ b/library/QATools/QATools/PageObject/PageFactory.php
@@ -16,6 +16,8 @@ use QATools\QATools\PageObject\Config\Config;
 use QATools\QATools\PageObject\Config\IConfig;
 use QATools\QATools\PageObject\Element\IElementContainer;
 use QATools\QATools\PageObject\ElementLocator\DefaultElementLocatorFactory;
+use QATools\QATools\PageObject\PageLocator\DefaultPageLocator;
+use QATools\QATools\PageObject\PageLocator\IPageLocator;
 use QATools\QATools\PageObject\PropertyDecorator\DefaultPropertyDecorator;
 use QATools\QATools\PageObject\PropertyDecorator\IPropertyDecorator;
 use QATools\QATools\PageObject\Url\IUrlFactory;
@@ -74,6 +76,13 @@ class PageFactory implements IPageFactory
 	protected $urlNormalizer;
 
 	/**
+	 * The page locator.
+	 *
+	 * @var IPageLocator
+	 */
+	protected $pageLocator;
+
+	/**
 	 * The current config.
 	 *
 	 * @var IConfig
@@ -96,6 +105,7 @@ class PageFactory implements IPageFactory
 		$this->setAnnotationManager($annotation_manager);
 		$this->setUrlFactory(new UrlFactory());
 		$this->setUrlNormalizer(new Normalizer($this->config->getOption('base_url')));
+		$this->setPageLocator(new DefaultPageLocator((array)$this->config->getOption('page_namespace_prefix')));
 	}
 
 	/**
@@ -160,6 +170,20 @@ class PageFactory implements IPageFactory
 	public function setUrlNormalizer(Normalizer $normalizer)
 	{
 		$this->urlNormalizer = $normalizer;
+
+		return $this;
+	}
+
+	/**
+	 * Sets the page locator.
+	 *
+	 * @param IPageLocator $page_locator The page locator.
+	 *
+	 * @return self
+	 */
+	public function setPageLocator(IPageLocator $page_locator)
+	{
+		$this->pageLocator = $page_locator;
 
 		return $this;
 	}
@@ -308,9 +332,9 @@ class PageFactory implements IPageFactory
 	 */
 	public function getPage($class_name)
 	{
-		$reflection = new \ReflectionClass($class_name);
+		$resolved_page_class = $this->pageLocator->resolvePage($class_name);
 
-		return $reflection->newInstanceArgs(array($this));
+		return new $resolved_page_class($this);
 	}
 
 }
--- a/library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
+++ b/library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
@@ -1,0 +1,147 @@
+<?php
+/**
+ * This file is part of the QA-Tools library.
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @copyright Alexander Obuhovich <aik.bold@gmail.com>
+ * @link      https://github.com/qa-tools/qa-tools
+ */
+
+namespace QATools\QATools\PageObject\PageLocator;
+
+
+use QATools\QATools\PageObject\Exception\PageFactoryException;
+
+/**
+ * Class to return fully qualified class names by its name.
+ *
+ * @method \Mockery\Expectation shouldReceive(string $name)
+ */
+class DefaultPageLocator implements IPageLocator
+{
+
+	/**
+	 * Prefix for page classes.
+	 *
+	 * @var array
+	 */
+	protected $namespacePrefixes = array();
+
+	/**
+	 * Creates the DefaultPageLocator.
+	 *
+	 * @param array $namespace_prefixes The page namespace prefixes.
+	 *
+	 * @throws PageFactoryException When no prefixes are given.
+	 */
+	public function __construct(array $namespace_prefixes)
+	{
+		if ( empty($namespace_prefixes) ) {
+			throw new PageFactoryException(
+				'No namespace prefixes passed',
+				PageFactoryException::TYPE_PAGE_MISSING_PREFIXES
+			);
+		}
+
+		foreach ( $namespace_prefixes as $prefix ) {
+			$this->namespacePrefixes[] = $this->normalize($prefix);
+		}
+	}
+
+	/**
+	 * Normalizes passed namespace.
+	 *
+	 * @param string $namespace Namespace.
+	 *
+	 * @return string
+	 */
+	protected function normalize($namespace)
+	{
+		$normalized_namespace = rtrim($namespace, '\\') . '\\';
+
+		if ( strpos($normalized_namespace, '\\') !== 0 ) {
+			$normalized_namespace = '\\' . $normalized_namespace;
+		}
+
+		return $normalized_namespace;
+	}
+
+	/**
+	 * Returns the fully qualified class name of a page by its name.
+	 *
+	 * @param string $name The name of the page.
+	 *
+	 * @return string
+	 * @throws PageFactoryException When no name is given.
+	 */
+	public function resolvePage($name)
+	{
+		if ( empty($name) ) {
+			throw new PageFactoryException('No page name given', PageFactoryException::TYPE_PAGE_NAME_MISSING);
+		}
+
+		$possible_pages = $this->buildPossiblePages($name);
+
+		return $this->getExistingPageClass($possible_pages);
+	}
+
+	/**
+	 * Builds all possible page classes from passed name and current prefixes.
+	 *
+	 * @param string $name Page name.
+	 *
+	 * @return array
+	 */
+	protected function buildPossiblePages($name)
+	{
+		$possible_classes = array();
+		$class_name = $this->buildClassNameFromName($name);
+
+		foreach ( $this->namespacePrefixes as $prefix ) {
+			$possible_classes[] = $prefix . $class_name;
+		}
+
+		return $possible_classes;
+	}
+
+	/**
+	 * Builds the class name from a given name by uppercasing the first letter of each word and removing the spaces.
+	 *
+	 * @param string $name The class name.
+	 *
+	 * @return string
+	 */
+	protected function buildClassNameFromName($name)
+	{
+		$class_name_parts = explode(' ', $name);
+
+		return count($class_name_parts) == 1 ? $name : implode('', array_map('ucfirst', $class_name_parts));
+	}
+
+	/**
+	 * Returns first existing class passed in array.
+	 *
+	 * @param array $possible_pages Possible page classes.
+	 *
+	 * @return string
+	 * @throws PageFactoryException When page class is not found.
+	 */
+	protected function getExistingPageClass(array $possible_pages)
+	{
+		foreach ( $possible_pages as $page_class ) {
+			if ( class_exists($page_class) ) {
+				return $page_class;
+			}
+		}
+
+		$message = sprintf(
+			'None of the possible classes were found: %s',
+			implode($possible_pages, ', ')
+		);
+
+		throw new PageFactoryException($message, PageFactoryException::TYPE_PAGE_CLASS_NOT_FOUND);
+	}
+
+}
+
--- a/tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
+++ b/tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
@@ -1,0 +1,143 @@
+<?php
+/**
+ * This file is part of the QA-Tools library.
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ *
+ * @copyright Alexander Obuhovich <aik.bold@gmail.com>
+ * @link      https://github.com/qa-tools/qa-tools
+ */
+
+namespace tests\QATools\QATools\PageObject\PageLocator;
+
+
+use Mockery as m;
+use QATools\QATools\PageObject\PageLocator\DefaultPageLocator;
+
+class DefaultPageLocatorTest extends \PHPUnit_Framework_TestCase
+{
+
+	/**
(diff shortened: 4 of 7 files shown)


library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @param string $namespace Namespace.
+	 *
+	 * @return string
+	 */
+	protected function normalize($namespace)
+	{
+		$normalized_namespace = trim($namespace, '\\') . '\\';
+
+		if ( strpos($normalized_namespace, '\\') !== 0 ) {
```
evangelion1204: In the end the namespace prefix should be ready to be prepended to the class itself. There should be always a `\` at the end to enable easy concatenation and a `\` at the front.

tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
```diff
@@ -0,0 +1,111 @@
...
+		return array(
+			array(
+				'ResolvePageMockedPrefixedClass',
+				array(),
+				'ResolvePageMockedPrefixedClass',
+			),
+			array(
+				'ResolvePageMockedPrefixedClassInNamespace',
+				array('\\Dummy', 'Example'),
+				'\\Example\\ResolvePageMockedPrefixedClassInNamespace',
```
evangelion1204: In this case only the fake class was just created in the `Example` namespace.
```php
m::mock($expected_resolved_page_class);
```

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -308,7 +332,9 @@ protected function getProperties(ISearchContext $search_context)
 	 */
 	public function getPage($class_name)
 	{
-		$reflection = new \ReflectionClass($class_name);
+		$resolved_page_class = $this->pageLocator->resolvePage($class_name);
+
+		$reflection = new \ReflectionClass($resolved_page_class);
```
evangelion1204: Done.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * Normalizes passed namespace.
+	 *
+	 * @param string $namespace Namespace.
+	 *
+	 * @return string
+	 */
+	protected function normalize($namespace)
```
evangelion1204: Done.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * Builds the class name from a given name by uppercasing the first letter of each word and removing the spaces.
+	 *
+	 * @param string $name The class name.
+	 *
+	 * @return string
+	 */
+	protected function buildClassNameFromName($name)
```
evangelion1204: Done.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,136 @@
...
+	{
+		$possible_pages = $this->buildPossiblePages($name);
+		$page_class = $this->findExistingPage($possible_pages);
+
+		if ( $page_class !== false ) {
+			return $page_class;
+		}
+
+		throw new PageFactoryException(
+			sprintf('"%s" was not found.', $name),
```
evangelion1204: Changed the exception to show all possible classes.

tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
aik099: I've looked into `buildPossiblePages` and saw, that it's checking if a given class already exists, then don't do any namespace prefixing. This might not be right because it allows for class from unregistered namespace to work.

What do you think?

evangelion1204: This  is right, in any case the class without namespace would win which is not desired, we should only check for the given namespaces. 
On the one hand we could always check for the default ```\\``` namespace or define a default if nothing is given.
```php
public function __construct(array $page_namespace_prefixes = array('\\'))
```

aik099:
Are you sure, that you've put right commit message into 0aba110040edb4efde3dcc4d27f0c685304638a0 commit? Now it says: `Adding assert for tag and type in typified elements.`, but looking at code it does something else.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
aik099: > `$normalized_namespace = trim($namespace, '\\') . '\\';`

This:

1. removes all `\` from both sides of string
2. adds `\` at the end

> `if ( strpos($normalized_namespace, '\\') !== 0 ) {`

This checks if `\` is found not at 1st symbol of a string, which is a bit unclear because we've just removed that leading `\`. I guess what you try to check here is if `\` is found in general by comparing to `false` instead.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,133 @@
...
+	 * Builds all possible page classes from passed name and current prefixes.
+	 *
+	 * @param string $name Page name.
+	 *
+	 * @return array
+	 */
+	protected function buildPossiblePages($name)
+	{
+		$class_name = $this->buildClassNameFromName($name);
+		$possible_classes = (array)$class_name;
```
aik099: The `$possible_classes = array($class_name);` would look more obvious because when `(array)` (typecasting) is used then this suggested that `$class_name` variable can be or be not an array in some cases.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,133 @@
...
+	 *
+	 * @param string $name The name of the page.
+	 *
+	 * @return string
+	 */
+	public function resolvePage($name)
+	{
+		$possible_pages = $this->buildPossiblePages($name);
+
+		return  $this->getExistingPageClass($possible_pages);
```
aik099: CS violation (2 spaces instead of 1 after `return`).

I recommend you:

1. updating PHP_CodeSniffer to latest version (2.1.0)
2. doing rebase on `develop` to get latest `composer.lock`
3. doing `composer install`

tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
aik099: Then:

1. this `array('\\')` should be default config value for `page_namespace_prefix` setting
2. don't add given class to list of possible class names
3. check (in tests) that when default Config object is used we're resolving classes from root namespace only (not sure what happens if given page name would be `a \ b`)
4. check (in tests) that with custom namespace list provided in the Config object we're looking for pages only in these namespaces

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * Returns first existing class passed in array.
+	 *
+	 * @param array $possible_pages Possible page classes.
+	 *
+	 * @return string|bool
+	 */
+	protected function findExistingPage(array $possible_pages)
```
aik099: Done.

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
evangelion1204: Indeed, I never thought about how this might be interpreted by someone who reads the code. Thanks for pointing this out :smile:

tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
evangelion1204: Thats what I am currently also unsure, with the given changes some tests fail, cause until now we used 
```php
protected $pageClass = '\\tests\\QATools\\QATools\\PageObject\\Fixture\\Page\\PageChild';
```
in PageFactoryTest. Which now results in double ```\``` in front, prefix is one and a ```\``` is included in the name. Maybe we should disallow absolute and only accept relative ones? Just stripping starting ```\``` is no option.

> check (in tests) that when default Config object is used we're resolving classes from root namespace only (not sure what happens if given page name would be a \ b)

I think if someone passed in ```a \ b``` we should search for this class in the defined namespaces:
```\prefix\a\b``` or ```\even\longer\prefix\a\b``` or ```\a\b```

File: library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php (hunk 1 above)
Comment (evangelion1204) [inline library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php]: I wanted to check if there is a starting ```\``` in case the global namespace was defined, else prepend it. In the end we could also use ```rtrim``` instead of ```trim``` and drop this check.

File: tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php (hunk 2 above)
Comment (aik099) [inline tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php]: > `protected $pageClass = '\\tests\\QATools\\QATools\\PageObject\\Fixture\\Page\\PageChild';`

In tests we can mock dummy page resolver that will return given page name (FQCN in this case) as it. Or set `\\tests\\QATools\\QATools\\PageObject\\Fixture\\Page` as namespace prefix and pass just `Page child`. In later case we maybe need to rename `$pageClass` into `$pageName` because it's no longer a class.

> I think if someone passed in a \ b we should search for this class in the defined namespaces:
\prefix\a\b or \even\longer\prefix\a\b or \a\b

Then proceed with original plan from https://github.com/qa-tools/qa-tools/pull/99/files#r22926901 comment.

And it turns out that `\` is also default namespace (unless user specified other one) in BehatPageObjectExtension as well.

File: library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php (hunk 1 above)
Comment (aik099) [inline library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php]: > I wanted to check if there is a starting \ in case the global namespace was defined, else prepend it. In the end we could also use rtrim instead of trim and drop this check.

In original BehatPageObject code there was `rtrim` indeed (https://github.com/sensiolabs/BehatPageObjectExtension/blob/master/src/SensioLabs/Behat/PageObjectExtension/PageObject/Factory/CamelcasedClassNameResolver.php#L128) and this check made sense. I think you should use rtrim.

> But what happens if a user passes a relative namespace as prefix. Maybe we also need a validation here.

This can't be handled unfortunately because in PHP you can't get current namespace in context where method was invoked to prepend it to given namespace. That's why in strings we always specify FQCN.

File: tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php (hunk 2 above)
Comment (evangelion1204) [inline tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php]: > check (in tests) that when default Config object is used we're resolving classes from root namespace only (not sure what happens if given page name would be a \ b)
> check (in tests) that with custom namespace list provided in the Config object we're looking for pages only in these namespaces

I just added this logic to the ```DefaultPageLocatorTest```. Or do you mean a test including the default ```Config``` object itself. Then I should add a test to the ```PageFactoryTest``` or just test ```Config``` as standalone, and its default values.

File: library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php [hunk 10]
```diff
@@ -0,0 +1,133 @@
...
+	 * @var array
+	 */
+	protected $pageNamespacePrefixes = array();
+
+	/**
+	 * Creates the DefaultPageLocator.
+	 *
+	 * @param array $page_namespace_prefixes The page namespace prefixes.
+	 */
+	public function __construct(array $page_namespace_prefixes = array('\\'))
```
aik099: Ha, no. Here we don't need a default value. The Config's class default setting for `page_namespace_prefix` should be `array('\\')` instead.

tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php
aik099: See https://github.com/qa-tools/qa-tools/pull/99/files#r23282552 . Here we only test that page resolution doesn't happen outside of given namespace list.

So we instantiate DefaultPageLocator with:
* root namespace (`\`)
* namespace with/without leading `\`
* namespace with/without trailing `\`
* namespace with `\` in the middle

and see how it does resolve given page name.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
evangelion1204: Ok, I already added this. But then should we check that something was passed in the constructor? The current code requires at least one namespace to be given or should we ignore this.

evangelion1204:
I think you need to help me with the Scrutinizer Failure. I don't know which one causes it.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
aik099: We can do this:
```php
if ( empty($page_namespace_prefixes) ) {
    throw new ....Exception('....');
}
``

Also I think that since class name is `*PageLocator` then using `page` word all over the page is too much. So maybe remove `page_` from `$page_namespace_prefixes` and `$pageNamespacePrefixes` class property.

**Comment by aik099** [comment]
The patches Scrutinizer CI suggests might be right, although I'm not sure if your PR introduced this: https://scrutinizer-ci.com/g/qa-tools/qa-tools/inspections/2b3be777-628d-482a-99df-ca934d124336/patches

> I think you need to help me with the Scrutinizer Failure. I don't know which one causes it.

I don't what issues Scrutinizer CI is talking about. I don't any new issues at all.

**Comment by aik099** [comment]
Just write when you're ready for next review.

File: tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php (hunk 2 above)
Comment (evangelion1204) [inline tests/QATools/QATools/PageObject/PageLocator/DefaultPageLocatorTest.php]: Added those additional tests.
Done with the next commit.

File: library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php (hunk 10 above)
Comment (evangelion1204) [inline library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php]: Done with the next commit.

**Comment by evangelion1204** [comment]
Ok, pushed, hopefully worked on all your comments else just let me know.

File: library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php [hunk 11]
```diff
@@ -0,0 +1,147 @@
...
+	 *
+	 * @param string $name The name of the page.
+	 *
+	 * @return string
+	 * @throws PageFactoryException When no name is given.
+	 */
+	public function resolvePage($name)
+	{
+		if ( empty($name) ) {
+			throw new PageFactoryException('No page name given.', PageFactoryException::TYPE_PAGE_NAME_MISSING);
```
aik099: No dots after exception messages please.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
```diff
@@ -0,0 +1,147 @@
...
+	protected function getExistingPageClass(array $possible_pages)
+	{
+		foreach ( $possible_pages as $page_class ) {
+			if ( class_exists($page_class) ) {
+				return $page_class;
+			}
+		}
+
+		$message = sprintf(
+			'None of the possible classes were found: %s.',
```
aik099: No dots after exception messages please.

library/QATools/QATools/PageObject/Config/Config.php
```diff
@@ -28,6 +28,7 @@ class Config implements IConfig
 	 */
 	protected $options = array(
 		'base_url' => '',
+		'page_namespace_prefix' => '\\',
```
aik099: Maybe we should have `array('\\')` here to indicate that multiple prefixes are also supported.

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/Config/Config.php
evangelion1204: Seems like a good idea, I will change that.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
evangelion1204: Sorry, I will keep this in mind the next time.

library/QATools/QATools/PageObject/PageLocator/DefaultPageLocator.php
evangelion1204: Done, part of next push.

evangelion1204:
Ok, rebased again and pushed. Ready for review and squash afterwards. In case that everything is fine.

aik099:
Nice. Please squash it then and I'll merge.

evangelion1204:
Ok, squashed. I can also rename the commit if you like :smile:

aik099:
Merged, thanks.

PR #113, opened 2015-01-06
Move the creation of the build folder to before_script on Travis

stof:
This is not a testing step but an initialization step


aik099:
Thanks.

PR #114, opened 2015-01-06
Refactor the Travis configuration to split the build doc

stof:
Building the doc to ensure it is valid is now in a dedicated job in the Travis matrix instead of being done in the jobs for each PHP version, as discussed in https://github.com/qa-tools/qa-tools/pull/96#discussion_r22526888

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,36 +7,39 @@ php:
     - 5.6
     - hhvm
 
+env:
+    global:
+        - BUILD_TYPE=code
+
+matrix:
+    include:
+        - php: 5.6 # A PHP version needs to be selected even if we won't use it there
+          env: BUILD_TYPE=doc
+
+before_install:
+    - if [ "$BUILD_TYPE" = "code" ] && [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then composer require satooshi/php-coveralls:dev-master --dev --no-update; fi;
+
+install:
+    - if [ "$BUILD_TYPE" = "code" ]; then composer update --prefer-source; fi;
+    - if [ "$BUILD_TYPE" = "doc" ]; then sudo pip install Sphinx sphinx_rtd_theme; fi;
+
 before_script:
     - export WEB_FIXTURES_HOST=http://localhost
     - export WEB_FIXTURES_BROWSER=firefox
-
-    - sh -e /etc/init.d/xvfb start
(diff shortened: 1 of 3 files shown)


aik099:
Maybe we can just add IF with "PHP 5.3" check before all Ruby-related commands. To it looks better then adding IF before each of commands.

stof:
Ruby ?

and IMO, it is easier to understand if the doc failures are in a job on its own. It is a totally different kind of failures

aik099:
What's concerning me is:
- now `.travis.yml` has IF statement on almost every line
- we might be (not sure if that's the case) getting less output logged from commands, that are executed in external files

.travis.yml
```diff
@@ -7,36 +7,39 @@ php:
...
-    - sudo easy_install -U Sphinx
-    - sudo pip install sphinx_rtd_theme
+    - if [ "$BUILD_TYPE" = "code" ]; then ./tests/run_selenium.sh; fi;
+    - if [ "$BUILD_TYPE" = "code" ]; then sudo ./tests/install_webserver.sh; fi;
 
 script:
     - mkdir -p build/logs
-    - ./vendor/bin/paratest --coverage-clover build/logs/clover.xml
-    - sphinx-build -nW -b html -d docs/build/doctrees docs docs/build/html
+    - if [ "$BUILD_TYPE" = "code" ]; then ./vendor/bin/paratest --coverage-clover build/logs/clover.xml; else echo "Nothing to do for this build"; fi;
```
aik099: Is `else` part necessary at all? There is no `else` part for other IF statements.

stof: this is just to make it clear that it is normal that the testing step is passing without running anything

stof:
@aik099 I opened (or +1ed on, I don't remember) a feature request on Travis about being able to use a totally different build setup (``before_install``, ``install``, ``before_script`` and ``script``) for some of the jobs in the matrix, but they rejected it saying they wanted to keep the travis.yml structure simple and there is a workaround already. Unfornately, both workarounds are making the config ugly:

- using IFs everywhere makes the config unreadable (and also the Travis output as commands which are not actually running still appear).
- moving everything in separate scripts means loosing the nice output folding and command display on Travis for each step

aik099:
No other options then. Everything is ready to be merged then?

aik099:
Merging in, thanks.

Issue #121, opened 2015-02-08
Add constraints for page creation


evangelion1204:
To detect that the current page does not match the current displayed website it would be great to add some constraints that are checked when the page is being created and throw an exception if those are not matched.

This will make it easier to write tests because we don't need to assert for the correct page class, which is not possible, because we don't expose the page locator to resolve the page class and do a assertion on the class instance.

A default constraint could be `@page-url`, which we could check for if it is matching. Although many shopping websites use SEO-friendly URLs and for those its not possible to detect the page by checking the URL.

We could introduce a `@constraint` annotation which could contain
* a URL
* an array of possible URLs 
* a RegExp to match against the URL
* a selector (like find-by) to an element which the page should contain - _I use CSS classes and data-attributes for JS to know which code to execute_
* query (GET) parameters of the current URL

aik099:
We can add `opened()` method, that would return true/false to indicate if this page is opened. Then you can have something `$this->assertEquals($this->page->opened())` in test code.

The most problematic case would be e-commerce websites where we can have `product detail page`, but url of that page looks like `/path/to/category/product-title.html`.

> a URL

We can use `WebAssert::addressEquals` method of Mink (see https://github.com/minkphp/Mink/blob/master/src/WebAssert.php#L49) as a start, which used in Behat scenario steps too.

The `cleanUrl` method (see https://github.com/minkphp/Mink/blob/master/src/WebAssert.php#L747) strips everything from compared url except SEO path and anchor. For websites, which pass current page name in query parameters this of course is not accepted solution.

> an array of possible URLs

Doesn't sound right. In product detail page example we'll post list of possible page urls. What if query parameters differ slightly and we already don't have full match.

> a RegExp to match against the URL

That is better, but yet again what about cases, where order of query string parameters is different. How do we compare then.

>  selector (like find-by) to an element which the page should contain - I use CSS classes and data-attributes for JS to know which code to execute

That sounds too website-specific. Having some markers on page to assist it's testing doesn't sound right.

> query (GET) parameters of the current URL

That sounds more like `@page-url` annotation itself where you can define url parts in any order and then we combine them to get final url. This way instead of comparing full url we can compare values of indicated query string parameters of it.

In any case we need to keep in mind 2 urls:
* seo: /language/path/to/template.html
* non-seo: /index.php?language=en&template=/path/to/template

aik099:
[...]

PR #123, opened 2015-02-22
Added "matches" method to "Page" class

### LINKED ISSUE #122: Add method to check for page instance by locator
To check if a given page is of name `Login Page` we need to implement a method in `Page` which uses the `PageLocator` of the `PageFactory` and resolves the class name of the given name.

evangelion1204:
Could also renamed to `assertClassName`, cause its currently an assertion and does not return anything.

Closes #122

### DIFF
--- a/library/QATools/QATools/PageObject/IPageFactory.php
+++ b/library/QATools/QATools/PageObject/IPageFactory.php
@@ -76,4 +76,13 @@ interface IPageFactory
 	 */
 	public function getPage($class_name);
 
+	/**
+	 * Fetches the FQCN of a page by given class name.
+	 *
+	 * @param string $class_name Page class name.
+	 *
+	 * @return string
+	 */
+	public function getPageClass($class_name);
+
 }
--- a/library/QATools/QATools/PageObject/IPageFactory.php
+++ b/library/QATools/QATools/PageObject/IPageFactory.php
@@ -68,21 +68,21 @@ interface IPageFactory
 	public function initElements(ISearchContext $search_context, IPropertyDecorator $property_decorator);
 
 	/**
-	 * Creates page by given class name.
+	 * Creates page by given name.
 	 *
-	 * @param string $class_name Page class name.
+	 * @param string $name Page name.
 	 *
 	 * @return Page
 	 */
-	public function getPage($class_name);
+	public function getPage($name);
 
 	/**
-	 * Fetches the FQCN of a page by given class name.
+	 * Fetches the FQCN of a page by given name.
 	 *
-	 * @param string $class_name Page class name.
+	 * @param string $name Page name.
 	 *
 	 * @return string
 	 */
-	public function getPageClass($class_name);
+	public function getPageClass($name);
 
 }
--- a/library/QATools/QATools/PageObject/Page.php
+++ b/library/QATools/QATools/PageObject/Page.php
@@ -107,25 +107,17 @@ abstract class Page extends DocumentElement implements ISearchContext
 	}
 
 	/**
-	 * Asserts that the page is of the given name.
+	 * Checks if page is of the given name.
 	 *
 	 * @param string $name Name of the page.
 	 *
-	 * @return self
-	 * @throws PageException When class name does not match.
+	 * @return bool
 	 */
-	public function assertPageName($name)
+	public function matches($name)
 	{
 		$class = $this->pageFactory->getPageClass($name);
 
-		if ( $this instanceof $class ) {
-			return $this;
-		}
-
-		throw new PageException(
-			'Page is not "' . $name . '": ' . get_class($this) . ' does not match ' . $class,
-			PageException::TYPE_PAGE_NOT_MATCHING
-		);
+		return $this instanceof $class;
 	}
 
 }
--- a/library/QATools/QATools/PageObject/Page.php
+++ b/library/QATools/QATools/PageObject/Page.php
@@ -109,22 +109,21 @@ abstract class Page extends DocumentElement implements ISearchContext
 	/**
 	 * Asserts that the page is of the given name.
 	 *
-	 * @param string $class_name Name of the page.
+	 * @param string $name Name of the page.
 	 *
 	 * @return self
-	 *
 	 * @throws PageException When class name does not match.
 	 */
-	public function matches($class_name)
+	public function assertPageName($name)
 	{
-		$class = $this->pageFactory->getPageClass($class_name);
+		$class = $this->pageFactory->getPageClass($name);
 
 		if ( $this instanceof $class ) {
 			return $this;
 		}
 
 		throw new PageException(
-			'Page is not "' . $class_name . '": ' . get_class($this) . ' does not match ' . $class,
+			'Page is not "' . $name . '": ' . get_class($this) . ' does not match ' . $class,
 			PageException::TYPE_PAGE_NOT_MATCHING
 		);
 	}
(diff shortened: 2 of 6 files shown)


library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +106,27 @@ public function setUrlBuilder(IBuilder $url_builder)
 		return $this;
 	}
 
+	/**
+	 * Asserts that the page is of the given name.
+	 *
+	 * @param string $class_name Name of the page.
+	 *
+	 * @return self
+	 *
```
aik099: No need for an empty line between `@return` and `@throws`.

library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +106,27 @@ public function setUrlBuilder(IBuilder $url_builder)
...
+	 * @throws PageException When class name does not match.
+	 */
+	public function matches($class_name)
```
aik099: The argument name seems to be wrong, because you give class as input and get class for comparison.

In fact it's wrong also for `PageFactory::getPage` method as well, because in the beginning it accepted page class names, but now it accepts page names which can also match page class name.

library/QATools/QATools/PageObject/IPageFactory.php
```diff
@@ -76,4 +76,13 @@ public function initElements(ISearchContext $search_context, IPropertyDecorator
...
 	public function getPage($class_name);
 
+	/**
+	 * Fetches the FQCN of a page by given class name.
+	 *
+	 * @param string $class_name Page class name.
+	 *
+	 * @return string
+	 */
+	public function getPageClass($class_name);
```
aik099: Here also argument should be `$name` I guess.

tests/QATools/QATools/PageObject/PageFactoryTest.php
```diff
@@ -263,6 +263,17 @@ public function testProxyFields()
 		$this->assertEquals('OK', $page->elementWithUse);
 	}
 
+	public function testgetPageClass()
```
aik099: Method should be named `testGetPageClass` (the `g` is capitalized).

tests/QATools/QATools/PageObject/PageFactoryTest.php
```diff
@@ -263,6 +263,17 @@ public function testProxyFields()
...
+	{
+		$factory = $this->createFactory(true, array('initPage', 'initElements', 'createDecorator'));
```
aik099: Any reason why we can't use factory initialized in `setUp` method?

tests/QATools/QATools/PageObject/PageTest.php
```diff
@@ -158,6 +158,27 @@ public function testOpenWithParamsCorrect()
...
+	{
+		$this->pageFactory->shouldReceive('getPageClass')->with('Default Page')->once()->andReturn($this->pageClass);
+
+		$this->page->matches('Default Page');
+
+		$this->assertTrue(true);
+	}
+
+	/**
+	 * @throws \QATools\QATools\PageObject\Exception\PageException
```
aik099: In test methods there is no point in specifying `@throws` at all.

aik099:
Today’s review completed.

tests/QATools/QATools/PageObject/PageTest.php
evangelion1204: Ok, just took the autocompletion, sorry.

library/QATools/QATools/PageObject/Page.php
evangelion1204: Done

library/QATools/QATools/PageObject/Page.php
evangelion1204: Renamed param to name.

library/QATools/QATools/PageObject/IPageFactory.php
evangelion1204: Also renamed to name.

tests/QATools/QATools/PageObject/PageFactoryTest.php
evangelion1204: Done.

tests/QATools/QATools/PageObject/PageTest.php
evangelion1204: Done.

tests/QATools/QATools/PageObject/PageFactoryTest.php
evangelion1204: I need a real factory and `$this->realFactory` fits my needs perfectly.

Done.

evangelion1204:
Pushed :smile: with all listed changes.

library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +106,26 @@ public function setUrlBuilder(IBuilder $url_builder)
...
+	 * @param string $name Name of the page.
+	 *
+	 * @return self
+	 * @throws PageException When class name does not match.
+	 */
+	public function assertPageName($name)
+	{
+		$class = $this->pageFactory->getPageClass($name);
+
+		if ( $this instanceof $class ) {
```
aik099: I'm not sure, that need to return something from this method, because asserting something is usually a last construct in test.

aik099: Having `assert*` methods in PageObject classes doesn't match PageObject pattern. It recommends for PageObject to provide all kind of info, but actual assertions needs to be made in tests.

aik099: Maybe we can really revert to `matches` method, that would just return true/false. Then we assert that result in the test code.

aik099:
Today’s review completed.

evangelion1204:
Ok, corrected and pushed.

library/QATools/QATools/PageObject/IPageFactory.php
```diff
@@ -68,12 +68,21 @@ public function initElementContainer(IElementContainer $element_container);
...
-	 * @param string $class_name Page class name.
+	 * @param string $name Page name.
 	 *
 	 * @return Page
 	 */
-	public function getPage($class_name);
+	public function getPage($name);
+
+	/**
+	 * Fetches the FQCN of a page by given name.
```
aik099: We've used full form of FQCN (the `fully qualified class name`) in other places. Maybe we should do the same here.

library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +106,18 @@ public function setUrlBuilder(IBuilder $url_builder)
 		return $this;
 	}
 
+	/**
+	 * Checks if page is of the given name.
+	 *
+	 * @param string $name Name of the page.
+	 *
+	 * @return bool
```
aik099: CS, should be `boolean` and PhpStorm shows that as error as well (thanks to PHP_CodeSniffer).

library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +106,18 @@ public function setUrlBuilder(IBuilder $url_builder)
...
+	 */
+	public function matches($name)
+	{
+		$class = $this->pageFactory->getPageClass($name);
+
+		return $this instanceof $class;
```
aik099: Usage of `instanceof` would give a false positive match, when given page name is parent page of current page:

```php
class BasePage {}
class PageA extends BasePage {}
class PageB extends BasePage {}

$page = new PageA();
$matches = $page->matches('Base page'); // this is true
```

We'd better use `get_class` here. And add tests to verify that comparison of pages with and without namespaces works fine.

aik099:
Today’s review completed.

aik099:
Agreed to close in favor of #121, because page class comparison only can detect mistakes in PageObject class internals, which doesn't really prove anything in the tests, where page url comparison would be more useful.

PR #124, opened 2015-03-04
Add page matcher

### LINKED ISSUE #121: Add constraints for page creation

evangelion1204:
First basic idea for the implementation of #121.

Closes #121


library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +113,38 @@ public function setUrlBuilder(IBuilder $url_builder)
...
+	 * @return bool
+	 */
+	public function opened()
+	{
+		if ( empty($this->urlMatches) ) {
+			return true;
+		}
+
+		foreach ( $this->urlMatches as $match ) {
+			if ( preg_match('#' . $match . '#', $this->getAbsoluteUrl()) ) {
```
aik099: I think, that `urlMatchers` array should contain full regexp with delimiters (the `#` in this case). Also the `#` is bad delimiter because we can match anchor as well.

library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +113,38 @@ public function setUrlBuilder(IBuilder $url_builder)
...
+
+	/**
+	 * Checks if the page is opened based on the specified url match.
+	 *
+	 * @return bool
+	 */
+	public function opened()
+	{
+		if ( empty($this->urlMatches) ) {
+			return true;
```
aik099: I'm not sure about this. Maybe if user haven't specified any, then we can't match or we must do a full url match.

aik099:
Current regexp matching has several drawbacks:
* we're comparing to full page url, that also includes domain and protocol and that forces each regexp either have that domain in it or have `$` at the end to compare only url end
* doing partial url comparison is very hard (e.g. anchor only, or presence of specific url parameter)

Since it's implementation of only 1 of 5 proposed matches, then I'll wait for others to be implemented.

library/QATools/QATools/PageObject/Page.php
evangelion1204: Indeed, maybe the annotation itself could contain different keys like `url` for exact match or `regexp` for a regular expression.

library/QATools/QATools/PageObject/Page.php
evangelion1204: A full url match might be the `page-url` annotation? 

Probably its better to throw an exception, cause there is nothing to match against specified.

evangelion1204:
> Since it's implementation of only 1 of 5 proposed matches, then I'll wait for others to be implemented.

Which of the matchers do we want to implement at all?
- complete url matching
- regexp url matching

I am not sure about the query check and the selector check is a bit to specific but I can implement it overriding `opened` in the inheriting `Page` class.

aik099:
> Which of the matchers do we want to implement at all?

I guess we need something, that would allow less writing to get desired match. For example something that will define url to be matched in a way `@page-url` annotation does: path, query params as array, anchor separately. Then if some of that is missing we consider that missing part shouldn't be compared.

evangelion1204:
Just added support for regexp and url as described above but I think I need to move `getMatches` although it just returns its data as an array.

In case we want a similar support like `@page-url` then I would suggest to extract the matching to a custom class, else there will be too much matching logic in `Page` itself.

aik099:
> In case we want a similar support like @page-url then I would suggest to extract the matching to a custom class, else there will be too much matching logic in Page itself.

If I remember correctly, then we discussed that idea before. However separate class won't have access to annotation data of the page unless page itself will query it first.

Maybe something like this would help:
```php
<?php

class PageMatcher implements IPageMatcher {

	protected $annotation_manager;

	public function __construct(AnnotationManager $annotation_manager)
	{
		....
	}

	public function matches(Page $page) // method of interface
	{
		// take whatever annotations needed or even lookup elements on a page
	}
}
```
Then as usual we:
1. set this class to factory
2. set it to page from initPage or from where we do "setUrlBuilder"
3. in `opened` method we call $this->pageMatcher->match($this)

evangelion1204:
> f I remember correctly, then we discussed that idea before. However separate class won't have access to annotation data of the page unless page itself will query it first.

You are correct and the proposed matcher looks really good.

evangelion1204:
Just added the anchor support to the current matcher. I will add a live unit test in this PR later.

In general it is now quite complicated to add a support for a new matching annotation.
Annotation should just store data which are used by different classes, but if the annotation itself would have the `match` function or is a factory to return a `matcher` instance which would simplify the whole process.
- create a new annotation with logic
or
- create new annotation
- create the matcher class which is returned by annotation

Now we need more steps, which in the end is clean :smile:
- create new annotation (if we require it)
- create a new `matcher` or extend existing one
- create the factory
- set matcher factory to page

aik099:
> Just added the anchor support to the current matcher.

You can't get page anchor value from PHP, so I wonder how you managed to compare it.

> create a new annotation with logic or

Nope for this, because annotation is data only.

> * create new annotation
> * create the matcher class which is returned by annotation

Maybe we can do this:

1. create several matcher classes, where each would done one matching type only (e.g. full url, partial url, regexp, etc.)
2. each matcher would also have associated annotation class, that won't be registred in factory by default
3. add `PageFactory::addMatcher(IPageMatcher $page_matcher)` method, that will accept matcher and call `->register($this->annotationManager)` on it so matcher can dynamically add new annotation classes that are required for him to function
4. maybe matchers needs to have priority (right inside them or given during registration) to allow custom matchers to be used before default ones

```php
interface IPageMatcher {

public function register(IAnnotationManager $annotation_manager);

public function match(Page $page);

}

class PageFactory
{
    protected $pageMatchers = array();

    public function addMatcher(IPageMatcher $page_matcher, $priority = 0)
        $page_matcher->register($this->annotationManager);
        $this->pageMatchers[] = array($page_matcher, $priority);
        // sort existing matchers according to their priority (matcher with higher number goes first)
    }

    public function matches(Page $page) {
        foreach($this->pageMatchers as $matcher_data) {
            if ( $matcher_data[0]->matches($page) ) {
                return true;
            }
        }

        return false;
    }

```

evangelion1204:
> You can't get page anchor value from PHP, so I wonder how you managed to compare it.

Hmm, I didn't gave it a try yet, I just use `getCurrentUrl()` of the session. I really don't know which drivers or any passed the current fragment.

> Nope for this, because annotation is data only.

As I guessed :smile: 

But the idea sounds great.

evangelion1204:
While adding the missing tests I found that after having called `Matcher::register` the `AnnotationManager` is fixed but the can be altered in the `PageFactory` which I wasn't aware. Either I could update/recreate the matchers or pass the `PageFactory` in the matchers to always have the current one.
I also removed the exception on missing annotations cause we default those 3 matchers.

Maybe its a good idea to make it configurable via our `Config`. Matchers can be defined by FQCN as an array.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -258,6 +284,54 @@ public function initPage(Page $page)
...
+	 * @param integer      $priority     Priority of the matcher.
+	 *
+	 * @return self
+	 */
+	public function addMatcher(IPageMatcher $page_matcher, $priority = 0)
+	{
+		$page_matcher->register($this->getAnnotationManager(), $this->getSession());
+		$this->pageMatchers[] = $page_matcher;
+
+		return $this;
```
aik099: No sorting by priority on new matcher addition yet, I guess.

aik099:
> While adding the missing tests I found that after having called Matcher::register the AnnotationManager is fixed but the can be altered in the PageFactory which I wasn't aware.

That is an interesting problem indeed. We have a lot of setters in PageFactory. Do we pass along what's been set to PageFactory to other objects inside the factory?

library/QATools/QATools/PageObject/PageFactory.php
evangelion1204: No, not yet, thats the next step. Which is no big deal with usort.

evangelion1204:
> That is an interesting problem indeed. We have a lot of setters in PageFactory. Do we pass along what's been set to PageFactory to other objects inside the factory?

The `AnnotationManager` is just used for the Decorator and the Property.
The `UrlBuilder` and `UrlNormilzer` is also fixed as soon as the page is created and not altered afterwards.

evangelion1204:
But maybe this is no problem for those cases, at this point they won't be changed, usually this stuff is changed on initialise.

evangelion1204:
Feel free to take a look, I added the lazy initialiser, which is not yet perfect.

library/QATools/QATools/PageObject/Annotation/PageUrlMatchAnnotation.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @var boolean|null
+	 */
+	public $secure;
+
+	/**
+	 * Match for anchor.
+	 *
+	 * @var string|null
+	 */
+	public $anchor;
```
aik099: Need to test check if anchor can be asserted at all. If not, then we'll just remove this one.

library/QATools/QATools/PageObject/Annotation/PageUrlMatchAnnotation.php
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * Initialize the annotation.
+	 *
+	 * @param array $properties Annotation parameters.
+	 *
+	 * @return void
+	 */
+	public function initAnnotation(array $properties)
+	{
+		$this->map($properties, array('url', 'regexp', 'secure', 'path', 'params', 'anchor'));
```
aik099: This is a bad idea. Since single annotation mix different location ways people would never guess that 3rd argument is secure for example.

I thought that there should separate annotations for radically different matching ways:
* full url match
* regexp match
* partial match

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+			if ( $scheme !== $parser->getComponent('scheme') ) {
+				return false;
+			}
+		}
+
+		if ( !empty($annotation->path) && $annotation->path !== $parser->getComponent('path') ) {
+			return false;
+		}
+
+		if ( !empty($annotation->params) ) {
```
aik099: With this we can't match to no url params (e.g. url without query string). Not sure if it's needed anywhere.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+			$params = array();
+			parse_str($parser->getComponent('query'), $params);
+
+			if ( $annotation->params !== $params ) {
```
aik099: That won't cover cases, where order if arguments in query string is different, than one in the matching annotation. I recommend doing something like `array_intersect` or similar to assert that both keys and values in both arrays match regardless of key order.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+				return false;
+			}
+		}
+
+		if ( !empty($annotation->anchor) && $annotation->anchor !== $parser->getComponent('fragment') ) {
```
aik099: If no way to get the anchor from PHP (not using Selenium-only features), then we'd better get rid of this one.

library/QATools/QATools/PageObject/Matcher/DefaultPageMatcherInitializer.php
```diff
@@ -0,0 +1,57 @@
...
+	/**
+	 * Initializes page matcher.
+	 *
+	 * @param IPageFactory $page_factory The page factory.
+	 *
+	 * @return self
+	 */
+	public function initialize(IPageFactory $page_factory)
+	{
+		foreach ( $this->matchers as $position => $matcher_class ) {
```
aik099: I presume we do sort matchers by position somewhere else.

library/QATools/QATools/PageObject/Matcher/FullUrlPageMatcher.php
```diff
@@ -0,0 +1,49 @@
...
+	 */
+	public function matches(Page $page)
+	{
+		/* @var $annotations PageUrlMatchAnnotation[] */
+		$annotations = $this->annotationManager->getClassAnnotations($page, '@page-url-match');
+
+		$url = $this->session->getCurrentUrl();
+
+		foreach ( $annotations as $annotation ) {
+			if ( !empty($annotation->url) && $url === $annotation->url ) {
```
aik099: Fact, that we compare url for `empty` here proves one more time, that 3 different matching annotations shouldn't be compared into single one.

library/QATools/QATools/PageObject/Matcher/RegExpUrlPageMatcher.php
```diff
@@ -0,0 +1,49 @@
...
+	 */
+	public function matches(Page $page)
+	{
+		/* @var $annotations PageUrlMatchAnnotation[] */
+		$annotations = $this->annotationManager->getClassAnnotations($page, '@page-url-match');
+
+		$url = $this->session->getCurrentUrl();
+
+		foreach ( $annotations as $annotation ) {
+			if ( !empty($annotation->regexp) && preg_match($annotation->regexp, $url) ) {
```
aik099: After separating `@page-match` annotation into 3 different ones we can get rid of `empty` part. In that case empty `regexp` would be an exception.

library/QATools/QATools/PageObject/Page.php
```diff
@@ -106,4 +106,14 @@ public function setUrlBuilder(IBuilder $url_builder)
...
 	}
 
+	/**
+	 * Checks if the page is opened based on the specified url match.
+	 *
+	 * @return boolean
+	 */
+	public function opened()
+	{
+		return $this->pageFactory->matches($this);
```
aik099: Method name read right. Looks like we're comparing page to factory.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -57,6 +60,7 @@ class PageFactory implements IPageFactory
 	protected $annotationRegistry = array(
 		'find-by' => '\\QATools\\QATools\\PageObject\\Annotation\\FindByAnnotation',
 		'page-url' => '\\QATools\\QATools\\PageObject\\Annotation\\PageUrlAnnotation',
+		'page-url-match' => '\\QATools\\QATools\\PageObject\\Annotation\\PageUrlMatchAnnotation',
```
aik099: And again:
- separate annotation class per matcher
- each matcher registers it's own in `register` method

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -241,6 +262,19 @@ public function getSession()
 	 */
 	public function initPage(Page $page)
 	{
+		return $this
+			->initPageUrl($page);
```
aik099: Is there ant reason for moving this into separate line?

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -258,6 +292,44 @@ public function initPage(Page $page)
...
 
 	/**
+	 * Registers new matcher.
+	 *
+	 * @param IPageMatcher $page_matcher The matcher.
+	 * @param integer      $priority     Priority of the matcher.
+	 *
+	 * @return self
+	 */
+	public function addMatcher(IPageMatcher $page_matcher, $priority = 0)
```
aik099: `MatcherRegistry` class sounds even better, when I look into this.

So, we:

1. create a MatcherRegistry, into which we add matchers
2. we set matcher registry to the factory
3. we call `$this->matcherRegistry->matches($page)

This is evolution of lazy initializer, because:
- it will have matcher objects or classes inside
- will query them for annotations to be registred upfront to avoid fatal error

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -258,6 +292,44 @@ public function initPage(Page $page)
...
+	{
+		$page_matcher->register($this->getAnnotationManager(), $this->getSession());
+		$this->pageMatchers[] = array($page_matcher, $priority);
+
+		usort($this->pageMatchers, function ($a, $b) {
+			return ($a[1] < $b[1]) ? -1 : 1;
```
aik099: What about `0`. If priorities match then we shouldn't move 2 records around. Priority isn't unique index in matcher array.

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/Annotation/PageUrlMatchAnnotation.php
evangelion1204: I added a live test and its possible. At least he passes :smile:

library/QATools/QATools/PageObject/Annotation/PageUrlMatchAnnotation.php
evangelion1204: You are right. I also thought about this and totally agree that there should be an annotation for each matcher.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: As mentioned above its possible, the live tests are working and showing that its possible to do at least with selenium.

library/QATools/QATools/PageObject/Matcher/DefaultPageMatcherInitializer.php
evangelion1204: If two matchers have the same priority they won't replace each other. The sorting is done in `addMatcher` with a `usort`.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: According to http://php.net/manual/en/language.operators.array.php `==` checks for same keys and values ignoring the order.

evangelion1204:
Just did a WIP commit, which already includes the `MatcherRegistry`, which is a nice improvement at all and separate annotations. I didn't alter all tests, just the live tests to see if its working in general. Yet I am not very satisfied with the registry and its initialise at all. 
As you mentioned we need to do this before the first page is being parsed, maybe I didn't do it elegant enough :wink: 

I will check it tomorrow again but also very thankful for your input.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @param AnnotationManager $annotation_manager The annotation manager.
+	 * @param Session           $session            The current mink session.
+	 *
+	 * @return self
+	 */
+	public function register(AnnotationManager $annotation_manager, Session $session)
+	{
+		parent::register($annotation_manager, $session);
+
+		$this->annotationManager->registry['match-component-url'] = '\\QATools\\QATools\\PageObject\\Annotation\\MatchComponentUrlAnnotation';
```
aik099: Maybe `match-url-component` would be better name.

library/QATools/QATools/PageObject/Annotation/MatchComponentUrlAnnotation.php
```diff
@@ -0,0 +1,66 @@
...
+
+
+use mindplay\annotations\Annotation;
+
+/**
+ * Annotation describing the component URL match.
+ *
+ * @usage('class'=>true, 'inherited'=>true, 'multiple'=>true)
+ */
+class MatchComponentUrlAnnotation extends Annotation
```
aik099: Let's:
1. introduce an `IUrlMatcherAnnotation` interface with `isValid` method
2. all 3 annotations would implement that method
3. each will check that minimal required set of annotation properties is filled-in to begin matching (e.g. regexp isn't empty for regexp matching annotation, etc.)

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	public function matches(Page $page)
+	{
+		/* @var $annotations MatchComponentUrlAnnotation[] */
+		$annotations = $this->annotationManager->getClassAnnotations($page, '@match-component-url');
+
+		$url = $this->session->getCurrentUrl();
+
+		foreach ( $annotations as $annotation ) {
+			if ( $this->matchComponent($annotation, $url) ) {
```
aik099: As 1st line in foreach we do `if ( !$annotation->isValid() ) { throw new .... }`.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @param MatchComponentUrlAnnotation $annotation The used annotation.
+	 * @param string                      $url        The current url.
+	 *
+	 * @return boolean
+	 */
+	protected function matchComponent(MatchComponentUrlAnnotation $annotation, $url)
+	{
+		$parser = new Parser($url);
+
+		if ( $annotation->secure === null
```
aik099: This IF would go to `IUrlMatcherAnnotation::isValid` method implementation.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Matches components.
+	 *
+	 * @param MatchComponentUrlAnnotation $annotation The used annotation.
+	 * @param string                      $url        The current url.
+	 *
+	 * @return boolean
+	 */
+	protected function matchComponent(MatchComponentUrlAnnotation $annotation, $url)
```
aik099: The Php Mess Detector says this method is very complex. Maybe we do something about it.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @var array
+	 */
+	protected $matchers = array();
+
+	/**
+	 * Current active page matchers.
+	 *
+	 * @var IPageMatcher[]|null
+	 */
+	protected $pageMatchers = null;
```
aik099: Why not an array only with empty array by default.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,120 @@
...
+	 * Initializes page matchers.
+	 *
+	 * @param AnnotationManager $annotation_manager Annotation manager.
+	 * @param Session           $session            Current session.
+	 *
+	 * @return self
+	 */
+	public function initialize(AnnotationManager $annotation_manager, Session $session)
+	{
+		if ( null !== $this->pageMatchers ) {
```
aik099: Instead you can compare lengths of `pageMatches` and `matchers` array and when they match, then it means the initialization was done.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -76,6 +78,20 @@ class PageFactory implements IPageFactory
...
+	 * @var MatcherRegistry
+	 */
+	protected $matcherRegistry;
+
+	/**
+	 * The registered page matchers.
+	 *
+	 * @var IPageMatcher[]
+	 */
+	protected $pageMatchers = array();
```
aik099: I thought that all machers now live in the `$matcherRegistry` property.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -241,6 +258,20 @@ public function getSession()
 	 */
 	public function initPage(Page $page)
 	{
+		$this->matcherRegistry->initialize($this->annotationManager, $this->_session);
```
aik099: The `initPage` happens on every page initialization. Why do we need to do initialization again and again for each page (I saw that `initialize` method just exists if initialization happened already)?

Let's think about why we're doing lazy initialization here in first place?

aik099:
Today’s review completed.

aik099:
Ups, hit the wrong button. Reopening.

aik099:
And you can most definitely squash these 16 commits :smile:

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Sure, sounds better, I already got rid of the page in front.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: To detect if everything is already initialised. In the end this probably does not matter, empty means not initialised, although no matchers might be specified.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Good point.

library/QATools/QATools/PageObject/PageFactory.php
evangelion1204: Didn't delete this one.

library/QATools/QATools/PageObject/PageFactory.php
evangelion1204: We can't do it too early, cause we might change the annotation manager, like done in the tests. The latest point to initialize the managers is directly before/in `initPage`, here we need the annotations added to the annotation manager.

Another way could to do the init everytime `setAnnotationManager` is called, we already do there a lot of stuff.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
aik099: Also an interesting scenario, when `page_matchers` is given in config, then they replace default ones. Maybe that's ok, I don't know.

library/QATools/QATools/PageObject/PageFactory.php
aik099: Do we allow to replace Mink's session used in factory? If we don't, then maybe we can change the code to prevent annotation manager changed as well?

evangelion1204: We do have a setter, but found ne reference that it is actually used from outside. For tests we solved it like this:
```php $this->pageFactory->shouldReceive('getSession')->andReturn($this->session);```
But then we force that we always use the getter even when being inside `PageFactory` else tests might not work or another way for DI is required, maybe via `Config`.



> Another way could to do the init everytime `setAnnotationManager` is called, we already do there a lot of stuff.
Totally forgot to mention `setMatcherRegistry`.

File: library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php (hunk 22 above)
Comment (evangelion1204) [inline library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php]: I am also not sure about that, we could make use the config to define additional matchers only, then we make sure our default ones are present. But it also might be that the user just wants his new one and then we don't need to load the defaults. 

In the end its just a matter of documentation where it is explained what to do if you want to create your custom matcher
* add default listeners (if required)
* add your custom listeners

File: library/QATools/QATools/PageObject/PageFactory.php (hunk 25 above)
Comment (aik099) [inline library/QATools/QATools/PageObject/PageFactory.php]: In fact all setter methods in `PageFactory` are only used once from constructor with default classes. What's currently in class constructor looks like a job for DIC, but we're not using one and we shouldn't force one on library users as well.

And we use both session and annotation manager, so we can't really move init code to `setAnnotationManager` as you proposed before.

File: library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php [hunk 26]
```diff
@@ -0,0 +1,120 @@
...
+	 * @var IPageMatcher[]|null
+	 */
+	protected $pageMatchers = null;
+
+	/**
+	 * Creates PageFactory instance.
+	 *
+	 * @param array $matchers Array of FQCN.
+	 */
+	public function __construct(array $matchers)
```
aik099: We need the constructor arguments to be `AnnotationManager $annotation_manager, Session $session` (same as initialize has currently). This way registry clearly states that he'll be needing those. In `PageFactory::__construct` we know both values and we can safely create this matcher registry. If user is manually creating it later then he can do what he wants.

aik099: Since we don't have `PageFactory::addMatcher` method, then we can safely initialize the matcher at time, when it's added to the registry (because registry now has annotation manager and session).

evangelion1204: But this still leaves the problem of how to get a custom annotation manager in for unit tests.

Maybe we need a builder pattern and make the factory itself completely immutable. Something like here in Java http://java.dzone.com/articles/immutability-with-builder-design-pattern

aik099: > But this still leaves the problem of how to get a custom annotation manager in for unit tests.

In tests where we set custom annotation manager we'd probably need to set the different matcher registry using that manager at same time. Just in idea.

> Maybe we need a builder pattern and make the factory itself completely immutable. Something like here in Java http://java.dzone.com/articles/immutability-with-builder-design-pattern

When argument count to PageFactory constructor was increased from 3 to 4 I've already switched to almost zero argument constructor and making all dependencies optional and populated through the constructor.

One of the alternatives to Builder pattern (according to http://en.wikipedia.org/wiki/Builder_pattern) is Factory method. Then implementation might be as follows:

* we remove all crazy setter methods (not just make them private/protected)
* add crazy amount of constructor arguments with typehints and null as default value
* constructor just sets given stuff
* add the static `PageFactory::createDefault` method that will:
** create all needed dependencies (what `__construct` does now)
** do the `return new static(......)

This way page factory will be immutable. We make constructor arguments optional to preserve BC for developers, that already instantiate it that way.

Still because that matcher factory that uses annotation manager this all looks more broken then before.

evangelion1204: > In tests where we set custom annotation manager we'd probably need to set the different matcher registry using that manager at same time. Just in idea.

You are right, totally forgot about this possibility.

I just remembered this type of builder pattern from my last meetup.



> Still because that matcher factory that uses annotation manager this all looks more broken then before.

Yes, it leaves the same problem :disappointed: Lets leave it like it is and for tests I can set the `MatcherRegistry` on my own.
Breaking the BC in this point it wouldn't be a too big deal, the init of the page factory usually happens just once, the work to get it running again would be minimal.

aik099: Please file separate task about that builder refactoring with ideas on how we can change that to make it more flexible. Maybe we'll have separate factory methods for common use cases.

One idea, that adds flexibility but reduces code clarity:

* allow any number of arguments in PageFactory constructor
* when processing 3+ argument (because 1st 2 remain the same) we look by class name of given object because they all are objects
* then we set what's given and create "new ..." for missing items

evangelion1204:
Ok, squashed and pushed. But I currently have no real idea how to test the sorting of `MatcherRegistry`.
Only idea I have is to use the fixtures of matchers and save the order of calling `match` or add `getMatchers`.

aik099:
> But I currently have no real idea how to test the sorting of MatcherRegistry.

You can create 2 matcher mocks with ordered expectation on `matches` method. This way it not only records that method was called, but also in the specific order.

evangelion1204:
With the current implementation I can't inject a mock directly, only a FQCN - thats why I made 2 fixtures.  
Can Mockery also handle this? I am not sure if it can be done with a named/alias mock.

aik099:
Maybe FQCN usage is wrong approach here. The easiest way would be adding matcher objects to registry and operating on them later. But that would prevent users from adding custom matchers via config.

Maybe while processing config we should check if it's an object already and use it and if it's not and object, then create a matcher from that class and add an object to registry.

evangelion1204:
Ok, did it like this, thats the same strategy I used for a lot of factories :smile: 

I changed the compare function and removed the check for equality cause it behaved in a strange way adding 2 matchers with the same priority. It always inverted the order:

* adding `a` with 0
* adding `b` with 0

results in [`b`, `a`] and not [`a`,`b`]

Maybe this is no problem, then I can add the equal check again.

aik099:
Are you sure, that during sorting records with identical priorities are swapped? I guess same happens in MySQL.

evangelion1204:
Yeah, tried it myself but in this case HHVM behaves correct, I think the sorting algorithm of  PHP is different, there is also a GIST: https://gist.github.com/dave1010/9274861 regarding this topic and this one http://stackoverflow.com/questions/12163532/php-usort-reorders-array-the-sort-value-is-the-same-for-all

aik099:
Don't bother with that. The note in the docs (I wish we had these docs) about that sorting would solve that issue.

evangelion1204:
Ok, just checked all your comments and everything should be done. 

Regarding Scrutinizer I find the suggestion to use a ternary operator instead of the clean if not that nice, but I can change it.

aik099:
> Regarding Scrutinizer I find the suggestion to use a ternary operator instead of the clean if not that nice, but I can change it.

Is it PhpAnalyzer suggestion or what PHP_CodeSniffer found? The PhpAnalyzer reported issues are not mandatory to be fixed by the way :)

evangelion1204:
But its always nice to see no new or at least as few as possible introduced issues ;)

aik099:
I'm unhappy with fact, that https://github.com/scrutinizer-ci/scrutinizer/issues/201 isn't resolved yet. Because of this we have a lot of errors like missing PhpDoc and such in tests folder.

evangelion1204:
Pushed with docs, not yet squashed.

evangelion1204:
Pushed and squashed.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
```diff
@@ -0,0 +1,79 @@
...
+
+
+use mindplay\annotations\Annotation;
+
+/**
+ * Annotation describing the component URL match.
+ *
+ * @usage('class'=>true, 'inherited'=>true, 'multiple'=>true)
+ */
+class UrlMatchComponentAnnotation extends Annotation implements IUrlMatcherAnnotation
```
aik099: Please also add support (here and in matcher class) for other component types: `user`, `pass` and `host `.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @return boolean
+	 */
+	protected function matchParams(UrlMatchComponentAnnotation $annotation, Parser $parser)
+	{
+		if ( $annotation->params !== null ) {
+			$params = array();
+			parse_str($parser->getComponent('query'), $params);
+
+			if ( $annotation->params != $params ) {
```
aik099: Please:

1. replace `$params` with `$parser->getParams()`
2. remove above 2 lines where `$params` variable is declared and used

tests/QATools/QATools/PageObject/Matcher/RegExpUrlPageMatcherTest.php
```diff
@@ -0,0 +1,93 @@
...
+use mindplay\annotations\AnnotationManager;
+use Mockery as m;
+use QATools\QATools\PageObject\Matcher\RegExpUrlPageMatcher;
+use QATools\QATools\PageObject\Page;
+use tests\QATools\QATools\TestCase;
+
+/**
+ * Class FullUrlPageMatcherTest
+ *
+ * @package tests\QATools\QATools\PageObject\Matcher
```
aik099: What is this? I haven't seen this in any other test. If there is no special purpose for that, then please remove.

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
evangelion1204: Done locally, also added support for port.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Done locally.

tests/QATools/QATools/PageObject/Matcher/RegExpUrlPageMatcherTest.php
evangelion1204: No idea, prob. added by the IDE. Dropped it.

evangelion1204:
Pushed :smile:

docs/annotations.rst
```diff
@@ -270,3 +270,51 @@ then it will override any default annotation, that might have been specified on
...
 ...
+
+@url-match-full
+---------------
+
+.. literalinclude:: examples/page_url_simple.php
+    :linenos:
+    :emphasize-lines: 6
+
+This annotation allows to check if a specific page is open comparing the full specified URL against the current open URL.
```
aik099: Please replace ` comparing the full specified URL against the current open URL.` with ` by comparing the specified full URL against the currently opened URL.`.

docs/annotations.rst
```diff
@@ -270,3 +270,51 @@ then it will override any default annotation, that might have been specified on
...
+
+@url-match-regexp
+-----------------
+This annotation allows to check if a specific page is open using a regular expression against the current open URL.
```
aik099: Please replace `is open using a regular expression against the current open URL.` with `is opened using a regular expression against the currently opened URL.`.

docs/annotations.rst
```diff
@@ -270,3 +270,51 @@ then it will override any default annotation, that might have been specified on
...
+
+.. code-block:: ruby
+
+    @url-match-full('/shoes\.html\?color=.+?$/')
+    @url-match-full('regexp' => '/shoes\.html\?color=.+?$/')
+
+@url-match-component
+--------------------
+This annotation allows to check if a specific page is open comparing different components of a URL against the current open URL.
```
aik099: Please replace ` is open comparing different components of a URL against the current open URL.` with ` is opened by comparing different components of the URL against the currently opened URL.`.

docs/annotations.rst
```diff
@@ -270,3 +270,51 @@ then it will override any default annotation, that might have been specified on
 @bem
 ----
 ...
+
+@url-match-full
```
aik099: Please reorder annotations on the page like this:

1. page-url
2. url-match-full
3. url-match-regexp
4. url-match-component
5. find-by
6. bem
7. element-name

This way we:
* first mention page-related annotation
* then mention element location annotations
* then mention all other annotations

docs/annotations.rst
```diff
@@ -270,3 +270,51 @@ then it will override any default annotation, that might have been specified on
...
+.. code-block:: ruby
+
+    @url-match-full('/shoes\.html\?color=.+?$/')
+    @url-match-full('regexp' => '/shoes\.html\?color=.+?$/')
+
+@url-match-component
+--------------------
+This annotation allows to check if a specific page is open comparing different components of a URL against the current open URL.
+
+Annotation can have 4 parameters, similar to ``@page-url``:
```
aik099: Please rephrase to avoid using exact numbers here. New parameters were added, but it's easy to forget to update places where parameter count it mentioned.

docs/annotations.rst
```diff
@@ -270,3 +270,51 @@ then it will override any default annotation, that might have been specified on
...
+
+#. ``path`` - specifies the path of an url
+#. ``params`` - specifies parameters in associative array format
+#. ``secure`` - specifies if secure connection is used
+#. ``anchor`` - specifies the fragment/anchor of an url
+#. ``host`` - specifies the host of an url
+#. ``port`` - specifies the port of an url
+#. ``user`` - specifies the user of an url
+#. ``pass`` - specifies the pass of an url
```
aik099: Please use `password` instead of `pass` when describing parameter.

library/QATools/QATools/PageObject/Annotation/IUrlMatcherAnnotation.php
```diff
@@ -0,0 +1,27 @@
...
+ * @link      https://github.com/qa-tools/qa-tools
+ */
+
+namespace QATools\QATools\PageObject\Annotation;
+
+
+/**
+ * Interface for matching annotations.
+ */
+interface IUrlMatcherAnnotation
```
aik099: The interface is called `IUrlMatcherAnnotation`, but all classes implementing are called `UrlMatch...` rather then `UrlMatcher...`. Maybe the interface name needs to be renamed to be inline with annotation classes, that implement it.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
```diff
@@ -0,0 +1,111 @@
...
+{
+
+	/**
+	 * Match for the path.
```
aik099: Not sure if we should mention `Match` word in here, because annotation itself doesn't do any matching. If you agree, then fix for this and other 2 annotations.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Validates required data.
+	 *
+	 * @return boolean
+	 */
+	public function isValid()
+	{
+		return $this->secure !== null
```
aik099: Any specific reason for comparing class properties not in the order of their declaration? If not then please either change class property declaration order or usage order in this method.

library/QATools/QATools/PageObject/Annotation/UrlMatchFullAnnotation.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Annotation describing the full URL match.
+ *
+ * @usage('class'=>true, 'inherited'=>true, 'multiple'=>true)
+ */
+class UrlMatchFullAnnotation extends Annotation implements IUrlMatcherAnnotation
+{
+
+	/**
+	 * Perfect match for the url.
```
aik099: What exactly `Perfect` word is doing in here? If you wanted to say that it exactly matches the url, then `Exact` term would be more appropriate.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,242 @@
...
+	 *
+	 * @param UrlMatchComponentAnnotation $annotation The used annotation.
+	 * @param string                      $url        The current url.
+	 *
+	 * @return boolean
+	 * @throws PageMatcherException When no matches specified.
+	 */
+	protected function matchComponent(UrlMatchComponentAnnotation $annotation, $url)
+	{
+		$parser = new Parser($url);
```
aik099: Please move below IF that does validation since it's used only if validation is passed.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
aik099: Was done in fact.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	/**
+	 * Matches the given page against the open.
+	 *
+	 * @param Page $page Page to match.
+	 *
+	 * @return boolean
+	 */
+	public function matches(Page $page)
```
aik099: The `matches` method is present in every matcher and is being used and yet isn't part of interface that every matcher implements.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,242 @@
...
+	{
+		/* @var $annotations UrlMatchComponentAnnotation[] */
+		$annotations = $this->annotationManager->getClassAnnotations($page, '@' . self::ANNOTATION);
+
+		$url = $this->session->getCurrentUrl();
```
aik099: 1. move out this into `$url` method parameter
2. add this to interface as well
3. let caller supply `$this->session->getCurrentUrl()` value

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,242 @@
...
+
+	/**
+	 * Initializes the Page Matcher.
+	 *
+	 * @param AnnotationManager $annotation_manager The annotation manager.
+	 * @param Session           $session            The current mink session.
+	 *
+	 * @return self
+	 */
+	public function register(AnnotationManager $annotation_manager, Session $session)
```
aik099: 1. remove `$session` parameter
2. also remove `AbstractPageMatcher::$session` property
3. don't pass session when `register` method is called

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,242 @@
...
+ * Responsible for matching given pages against displayed page.
+ *
+ * @method \Mockery\Expectation shouldReceive(string $name)
+ */
+class ComponentUrlPageMatcher extends AbstractPageMatcher
+{
+	const ANNOTATION = 'url-match-component';
+
+	/**
+	 * Initializes the Page Matcher.
```
aik099: The method comment is radically incorrect, because the constructor initializes the class, not some method.

Any objections against supplying the AnnotationManager in the constructor for matchers instead of having the `register` method?

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+	protected $pageMatchers = array();
+
+	/**
+	 * Creates PageFactory instance.
+	 *
+	 * @param AnnotationManager $annotation_manager Annotation manager.
+	 * @param Session           $session            Current session.
+	 * @param array             $matchers           Array of FQCN of matchers.
+	 */
+	public function __construct(AnnotationManager $annotation_manager, Session $session, array $matchers = array())
```
aik099: Please:
1. remove `$session` parameter
2. remove `$matchers` parameter and code, that processes it

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * Current active page matchers.
+	 *
+	 * @var IPageMatcher[]
+	 */
+	protected $pageMatchers = array();
+
+	/**
+	 * Creates PageFactory instance.
```
aik099: We're creating matcher registry, not PageFactory here :wink: .

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @var AnnotationManager
+	 */
+	protected $annotationManager;
+
+	/**
+	 * The current config.
+	 *
+	 * @var array
+	 */
+	protected $matchers = array();
```
aik099: Please remove this.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * Current active page matchers.
+	 *
+	 * @var IPageMatcher[]
+	 */
+	protected $pageMatchers = array();
```
aik099: Please rename into `$matchers`.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * Registers new matcher.
+	 *
+	 * @param string  $page_matcher The matcher FQCN.
+	 * @param integer $priority     Priority of the matcher.
+	 *
+	 * @return self
+	 */
+	public function registerMatcher($page_matcher, $priority = 0)
```
aik099: Please:

1. rename `registerMatcher` into `add`
2. add `IPageMatcher` type hint for `$page_matcher` parameter

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @return self
+	 */
+	public function initialize()
+	{
+		if ( count($this->matchers) === count($this->pageMatchers) ) {
+			return $this;
+		}
+
+		usort($this->matchers, function ($a, $b) {
+			if ( $a[1] === $b[1] ) {
```
aik099: Please replace whole sorting callback body with `return strcmp($a[1], $b[1]);` if you doing ascending sorting here.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+	 * Matches the page against registered matchers.
+	 *
+	 * @param Page $page Page to match.
+	 *
+	 * @return boolean
+	 */
+	public function match(Page $page)
+	{
+		foreach ( $this->pageMatchers as $matcher ) {
+			if ( $matcher->matches($page) ) {
```
aik099: Pass session in 2nd parameter.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * Matches the page against registered matchers.
+	 *
+	 * @param Page $page Page to match.
+	 *
+	 * @return boolean
+	 */
+	public function match(Page $page)
```
aik099: Add session as 2nd parameter.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -106,6 +114,11 @@ public function __construct(Session $session, IConfig $config = null)
 		$this->setUrlFactory(new UrlFactory());
 		$this->setUrlNormalizer(new Normalizer($this->config->getOption('base_url')));
 		$this->setPageLocator(new DefaultPageLocator((array)$this->config->getOption('page_namespace_prefix')));
+		$this->setMatcherRegistry(
+			new MatcherRegistry($annotation_manager, $session, $this->config->getOption('page_matchers'))
+		);
+
+		$this->matcherRegistry->initialize();
```
aik099: Please remove this, because we're adding matcher objects now.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -106,6 +114,11 @@ public function __construct(Session $session, IConfig $config = null)
 		$this->setUrlFactory(new UrlFactory());
 		$this->setUrlNormalizer(new Normalizer($this->config->getOption('base_url')));
 		$this->setPageLocator(new DefaultPageLocator((array)$this->config->getOption('page_namespace_prefix')));
+		$this->setMatcherRegistry(
+			new MatcherRegistry($annotation_manager, $session, $this->config->getOption('page_matchers'))
```
aik099: 1. move out MatcherRegistry class creation into `$matcher_registry` temp variable
2. remove `$session` passing
3. remove matcher passing
4. iterate over config option value and:
 * create classes from given names
 * call `$matcher_registry->add(new $matcher_class())`
5. do `$this->setMatcherRegistry($matcher_registry)`

aik099:
Today’s review completed. Sorry for huge comment amount, but after some time I was able to take fresh look so to say on the code and spotted some issues that weren't obvious before.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
aik099: Also interesting fact:

1. priority usage/sorting makes no sense because we're not enabling all matchers at one, but only use ones, that user have specified via Page class annotations
2. each matcher has 1 annotation only

Maybe knowing that we can:

* use matcher annotation name as key during it's registration
* let matcher registry query annotations and provide them to matcher (this way matcher won't be dependent on annotation manager)

docs/annotations.rst
evangelion1204: Done locally.

docs/annotations.rst
evangelion1204: Done locally.

docs/annotations.rst
evangelion1204: Done locally.

docs/annotations.rst
evangelion1204: Makes sense, I will also correct @page-url

docs/annotations.rst
evangelion1204: Done locally.

docs/annotations.rst
evangelion1204: > If we support placing several url-match-* annotations to act as OR, then please write about this as well.

Good point, also added tests for this case.

library/QATools/QATools/PageObject/Annotation/IUrlMatcherAnnotation.php
evangelion1204: Indeed, correcting.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
evangelion1204: I think it should stay here, cause it is the annotation for the URL component matcher.

Normally we use the name of the annotation camel cased with `Annotation` appended. Here `@url-match-component` to `UrlMatchComponentAnnotation`.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
evangelion1204: No reason, but I will rearrange them.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Its part of `IPageMatcher`.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Good point, this makes the whole code easier to test and understandable.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally :blush:

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/PageFactory.php
evangelion1204: I will do the registerAnnotation in the `MatcherRegistry` or should we also extract it, in this case we don't even need to pass the annotation manager to the registry.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: Done locally.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
evangelion1204: > rename registerMatcher into add
> add IPageMatcher type hint for $page_matcher parameter
> sort matchers by priority right in this method (adding new matcher would resort all of them)

Done locally.



> let matcher registry query annotations and provide them to matcher (this way matcher won't be dependent on annotation manager)

Either we ask for the constant (when adding an annotation) to fetch the annotation, but this will only work as long as we have just one per matcher.
Or did I got it wrong?

library/QATools/QATools/PageObject/PageFactory.php
aik099: You can try extracting it and if it will look better.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
aik099: We can only have 1 annotation per matcher. This one annotation can be as complex as needed. For now I can't imagine case, where one matcher need to operate on 2+ different annotations to figure out how matching will happen.

library/QATools/QATools/PageObject/PageFactory.php
evangelion1204: Just took a look and thought about it, I think we should keep it as it is now.

aik099: Maybe later we can transform page factory into something plugin-based like Behat 3 did (compared to Behat 2).

evangelion1204:
Ok, done, pushing corrections.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * @return self
+	 */
+	public function add(IPageMatcher $page_matcher, $priority = 0)
+	{
+		$page_matcher->registerAnnotations($this->annotationManager);
+
+		$this->matchers[] = array('instance' => $page_matcher, 'priority' => $priority);
+
+		usort($this->matchers, function ($a, $b) {
```
aik099: Please rename `$a` into `$matcher_a` and `$b` into `$matcher_b`.

---

This is reported as CS error in PhpStorm.

docs/annotations.rst
aik099: Seems, that you forgot to push this one, because it still says `pass` here and in `UrlMatchComponentAnnotation` class.

library/QATools/QATools/PageObject/Annotation/UrlMatchComponentAnnotation.php
aik099: I'm not talking about annotation class name, but about DocBlock for it's properties. Since the annotation is data only (no processing logic assumed), then it just holds path. It's the matcher class, that decides to use data path for matching.

Same goes for other matcher-related annotation property DocBlocks.

library/QATools/QATools/PageObject/Annotation/UrlMatchFullAnnotation.php
aik099: Done.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
aik099: OK, that was me misinterpreting override icon in PhpStorm.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
aik099: This is not addressed.

library/QATools/QATools/PageObject/PageFactory.php
aik099: Done.

library/QATools/QATools/PageObject/Matcher/IPageMatcher.php
```diff
@@ -0,0 +1,45 @@
...
+use Behat\Mink\Session;
+use mindplay\annotations\AnnotationManager;
+use QATools\QATools\PageObject\Page;
+
+/**
+ * Responsible for matching given pages against displayed page.
+ *
+ * @method \Mockery\Expectation shouldReceive(string $name)
+ */
+interface IPageMatcher
```
aik099: Please add `getAnnotationName` method for each matcher to implement. Existing solution with class constant doesn't really force developer specify annotation name, but proposed solution will.

library/QATools/QATools/PageObject/Matcher/AbstractPageMatcher.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+abstract class AbstractPageMatcher implements IPageMatcher
+{
+
+	/**
+	 * The annotation manager required to read annotations of the given page.
+	 *
+	 * @var AnnotationManager
+	 */
+	protected $annotationManager;
```
aik099: Please remove this, because we decided not to store it inside the matcher.

library/QATools/QATools/PageObject/Matcher/AbstractPageMatcher.php
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Registers annotations, used by matcher.
+	 *
+	 * @param AnnotationManager $annotation_manager The annotation manager.
+	 *
+	 * @return self
+	 */
+	public function registerAnnotations(AnnotationManager $annotation_manager)
```
aik099: Please remove this, because we decided not to store it inside the matcher.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,240 @@
...
+	/**
+	 * Registers annotations, used by matcher.
+	 *
+	 * @param AnnotationManager $annotation_manager The annotation manager.
+	 *
+	 * @return self
+	 */
+	public function registerAnnotations(AnnotationManager $annotation_manager)
+	{
+		parent::registerAnnotations($annotation_manager);
```
aik099: One method in parent class will be removed this line should also be removed, because we can't store annotation manager in method, that is just supposed to use it once and leave it.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,240 @@
...
+use QATools\QATools\PageObject\Url\Parser;
+
+/**
+ * Responsible for matching given pages against displayed page.
+ *
+ * @method \Mockery\Expectation shouldReceive(string $name)
+ */
+class ComponentUrlPageMatcher extends AbstractPageMatcher
+{
+	const ANNOTATION = 'url-match-component';
```
aik099: This, according to another comment, will be replaced with `getAnnotationName` method.

library/QATools/QATools/PageObject/Matcher/ComponentUrlPageMatcher.php
```diff
@@ -0,0 +1,240 @@
...
+	 *
+	 * @param Page   $page Page to match.
+	 * @param string $url  The URL.
+	 *
+	 * @return boolean
+	 */
+	public function matches(Page $page, $url)
+	{
+		/* @var $annotations UrlMatchComponentAnnotation[] */
+		$annotations = $this->annotationManager->getClassAnnotations($page, '@' . self::ANNOTATION);
```
aik099: This is tricky, but please bear with me:

1. move this line up to the `MatcherRegistry`
2. the matcher registry would call `getAnnotationName` on the matcher class and then get all needed annotations
3. replace `$page` argument with `array $annotations` argument
4. make the `$url` parameter first one, because we're matching url to the annotations, not the annotations to the url

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/Matcher/AbstractPageMatcher.php
evangelion1204: Its clear that we won't pass the manager anymore, but where do we register the annotations itself? The `getAnnotationName` is just the used name and not the matched FQCN to the given name. It prob. makes sense to also add a `getAnnotationClass` to the matcher too.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -337,4 +370,36 @@ public function getPage($class_name)
 		return new $resolved_page_class($this);
 	}
 
+	/**
+	 * Builds the matcher registry.
+	 *
+	 * @param AnnotationManager $annotation_manager The annotation manager.
```
aik099: Please remove this line, because method no longer have any parameters.

library/QATools/QATools/PageObject/PageFactory.php
```diff
@@ -337,4 +370,36 @@ public function getPage($class_name)
...
+	 *
+	 * @return MatcherRegistry
+	 */
+	protected function buildMatcherRegistry()
+	{
+		$matcher_registry = new MatcherRegistry($this->annotationManager);
+
+		foreach ( $this->config->getOption('page_matchers') as $index => $matcher_class ) {
+			$matcher_registry->add(new $matcher_class, $index);
```
aik099: Please replace `new $matcher_class` with `new $matcher_class()`.

library/QATools/QATools/PageObject/Matcher/MatcherRegistry.php
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * Adds a page matcher instance.
+	 *
+	 * @param IPageMatcher $page_matcher The page matcher instance.
+	 * @param integer      $priority     Priority of the matcher.
+	 *
+	 * @return self
+	 */
+	public function add(IPageMatcher $page_matcher, $priority = 0)
```
aik099: Actually please use approach I've used in another library:

1. the `IPageMatcher::getPriority` method that would return unique priority for each matcher
2. when adding matcher take priority from calling `$page_matcher->getPriority()`
3. use priority as a key in `$this->matchers` array
4. if priority isn't unique (compared to other matchers from `$this->matchers` array) then throw an exception
5. instead of `usort` below actually just sort array by keys

library/QATools/QATools/PageObject/Matcher/AbstractPageMatcher.php
aik099: Yes, that makes sense. After these changes the `IPageMatcher::registerAnnotations` doesn't really do anything, because with `IPageMatcher::getAnnotationName` and `IPageMatcher::getAnnotationClass` the `MatcherRegistry` class can register annotations itself right in the `MatcherRegistry::add` method.


I confirm, that `getAnnotationName` and `getAnnotation` class methods

aik099:
Today’s review completed.

aik099:
I'm a bit busy at work currently. Will try to review your changes as soon as I can. Apologies for unexpected delays.

aik099:
Closing in favor of #188, where all your changes were incorporated. Thank you, @evangelion1204 .

PR #125, opened 2015-03-05
Refactor the Travis config for different jobs

stof:
Instead of checking the build type in each command, the new config takes advantage of an undocumented Travis feature which allows to overwrite the whole config for a given job in the matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,32 +14,36 @@ env:
 matrix:
     include:
         - php: 5.6 # A PHP version needs to be selected even if we won't use it there
-          env: BUILD_TYPE=doc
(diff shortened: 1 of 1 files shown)


stof:
This can become even better by switching to the Python VM for the doc build. See the second commit

aik099:
> takes advantage of an undocumented Travis feature 

Hopefully this feature won't disappear in future and all will stop working because of that.

> This can become even better by switching to the Python VM for the doc build.

The log is very impressive.

Ready to merge?

stof:
It is (except that the target branch is totally outdated)

aik099:
Merging, thanks.

stof:
@aik099 I asked the Travis team whether it is something they will preserve in the future

PR #129, opened 2015-04-03
Added support for ports

### LINKED ISSUE #127: Add support for port definition
Currently it is not possible to add a port to the base url. Some users tend to run a specific html-server for example on port 8080 (default port of node `http-server`).

evangelion1204:
Closes #127

### DIFF
--- a/docs/annotations.rst
+++ b/docs/annotations.rst
@@ -227,6 +227,16 @@ By specifying ``secure`` annotation parameter (3rd) value it's possible to force
     @page-url('url' => 'products/shoes.html', 'secure' => false)
     @page-url('url' => 'https://www.example.com/products/shoes.html', 'secure' => false)
 
+Custom ports
+^^^^^^^^^^^^
+Like for `base_url` it is possible to include a port in an absolute URL:
+
+.. code-block:: ruby
+
+    @page-url('http://www.example.com:8080/products/shoes.html')
+
+
+
 @timeout
 --------
 This annotation is used to specify maximum waiting time (in seconds), after which search attempt for an element, that
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -54,6 +54,8 @@ The following configuration options are available:
 
 * ``base_url`` - allows to specify Base URL to be used to transform all relative urls from :ref:`@page-url annotations <page-url>` into absolute urls.
 
+If port is specified as part of `base_url` then it will be used in every built url unless specified exactly in the `@page-url` annotation.
+
 Connecting to Behat
 -------------------
 .. note:: Throughout this tutorial it's assumed that working Behat with MinkExtension is configured and connected to a project that needs to be tested.
--- a/library/QATools/QATools/PageObject/Url/Builder.php
+++ b/library/QATools/QATools/PageObject/Url/Builder.php
@@ -50,6 +50,13 @@ class Builder implements IBuilder
 	protected $host = '';
 
 	/**
+	 * The port of the url.
+	 *
+	 * @var integer|null
+	 */
+	protected $port = null;
+
+	/**
 	 * GET url parameters.
 	 *
 	 * @var array
@@ -73,6 +80,7 @@ class Builder implements IBuilder
 		$this->host = $components['host'];
 		$this->protocol = $components['scheme'];
 
+		$this->port = !empty($components['port']) ? $components['port'] : '';
 		$this->anchor = !empty($components['fragment']) ? $components['fragment'] : '';
 
 		if ( !empty($components['query']) ) {
@@ -89,7 +97,7 @@ class Builder implements IBuilder
 	 */
 	public function build(array $params = array())
 	{
-		$final_url = $this->getProtocol() . '://' . $this->getHost() . $this->getPath();
+		$final_url = $this->getProtocol() . '://' . $this->getHost() . $this->getPortForBuild() . $this->getPath();
 		$final_params = array_merge($this->getParams(), $params);
 
 		if ( !empty($final_params) ) {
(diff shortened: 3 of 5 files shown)


library/QATools/QATools/PageObject/Url/Builder.php
```diff
@@ -20,6 +20,9 @@
  */
 class Builder implements IBuilder
 {
+	const DEFAULT_PORT_HTTP = '80';
```
aik099: Any reason for storing numbers as string?

tests/QATools/QATools/PageObject/Url/BuilderTest.php
```diff
@@ -140,6 +142,36 @@ public function buildDataProvider()
 				array(),
 				'http://domain.tld/path?param=value#anchor',
 			),
+			array(
+				array(
+					'scheme' => 'http',
+					'host' => 'domain.tld',
+					'path' => '/path',
+					'port' => '8080',
```
aik099: Any reason for specifying port as string?

library/QATools/QATools/PageObject/Url/Builder.php
evangelion1204: Ok, thats true, in this case I will remove them.

tests/QATools/QATools/PageObject/Url/BuilderTest.php
evangelion1204: No particular, its true that they should be a number, I will change it.

evangelion1204:
Ok, corrected.

evangelion1204:
And the other stuff :smile: sorry.

library/QATools/QATools/PageObject/Url/Builder.php
```diff
@@ -104,6 +112,26 @@ public function build(array $params = array())
...
+	 */
+	protected function getPortForBuild()
+	{
+		if ( !$this->getPort() ) {
+			return '';
+		}
+
+		$default_ports = array('http' => 80, 'https' => 443);
+
+		if ( $default_ports[$this->getProtocol()] == $this->getPort() ) {
```
aik099: No Yoda Conditions please. The constant part should be on the right of comparison.

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/Url/Builder.php
evangelion1204: Corrected and pushed.

evangelion1204:
Pushed.

aik099:
Totally forgot that we have docs for annotations, so please describe port usage there as well.

library/QATools/QATools/PageObject/Url/Builder.php
```diff
@@ -104,6 +112,26 @@ public function build(array $params = array())
...
+	 */
+	protected function getPortForBuild()
+	{
+		if ( !$this->getPort() ) {
+			return '';
+		}
+
+		$default_ports = array('http' => 80, 'https' => 443);
+
+		if ( $this->getPort() == $default_ports[$this->getProtocol()] ) {
```
aik099: Use `===` here.

evangelion1204:
I need to add the support in the annotations as well.

evangelion1204:
Ok, corrected `===` and added a doc entry. But for all use cases that come to my mind I am not sure if we should add `port` to `@page-url` if somebody uses a custom port and secure he might need to switch the port.
* 8080 => http
* 8081 => https

aik099:
Since the port is empty by default, then no matter if secure is used or not or it's value we're not adding port to url. If developer specifies both secure & port, then he should know that given port would be used regardless the protocol.

docs/annotations.rst
```diff
@@ -227,6 +227,16 @@ By specifying ``secure`` annotation parameter (3rd) value it's possible to force
     @page-url('url' => 'products/shoes.html', 'secure' => false)
     @page-url('url' => 'https://www.example.com/products/shoes.html', 'secure' => false)
 
+Custom ports
+^^^^^^^^^^^^^^^^^^^^
```
aik099: The `^` symbol count must match above line length in symbols. The PhpStorm reports this. I wonder why this haven't failed documentation build.

docs/getting-started.rst
```diff
@@ -54,6 +54,8 @@ The following configuration options are available:
 
 * ``base_url`` - allows to specify Base URL to be used to transform all relative urls from :ref:`@page-url annotations <page-url>` into absolute urls.
 
+.. note:: It is also possible to use a specific port like ``8080`` via ``http://example.com:8080``.
```
aik099: We have a lot more options, then `base_url`, but they are missing in docs current. When we added then, then `note` you've added can't be easily matched to `base_url` setting. Maybe add this in 2nd sentence after `base_url` option is described.

Text like this sounds better to me (change as you see fit):
```
If port is specified as part of `base_url` then it will be used in every built url unless specified exactly in the `@page-url` annotation.
```

evangelion1204: True, then I can drop the note itself, most people should be familiar specifying a port in a url but should still keep the custom port in the annotation, although its basic knowledge.

evangelion1204:
Took you sentence and removed the info, I like it more and tweaked the text in custom ports of `@find-by`.

tests/QATools/QATools/PageObject/Url/ParserTest.php
```diff
@@ -68,6 +70,7 @@ public function testGetComponentDefault()
 		$this->assertEquals('default_path', $url_parser->getComponent('path', 'default_path'));
 		$this->assertEquals('default_query', $url_parser->getComponent('query', 'default_query'));
 		$this->assertEquals('default_fragment', $url_parser->getComponent('fragment', 'default_fragment'));
+		$this->assertEquals(80, $url_parser->getComponent('port', 80));
```
aik099: Better change port to something non-default here to indicate that what's given in 2nd parameter to `getComponent` is returned.

aik099:
You're almost there. Just one more change.

evangelion1204:
Ok pushed with port `8080` instead of `80`

aik099:
Good work. Merging. Thanks.

PR #138, opened 2015-05-31
Property forwarding support in proxies

### LINKED ISSUE #137: Proxies does not work for properties
Having an proxied element like
```php
class LoginPane extends AbstractElementContainer
{

	/**
	 * The e-mail.
	 *
	 * @var TextInput
	 * @find-by('id' => 'some-id')
	 */
	public $email;

}
```

And trying to access it by:

```php
$this->loginPane->email->click();
```

Will result in ` Notice: Undefined property: QATools\QATools\PageObject\Proxy\WebElementProxy::$email in ...`

We should also proxy the properties implementing `__get` and `__set`.

evangelion1204:
Implemented support property forwarding in proxies.

Closes #137

### DIFF
--- a/library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
+++ b/library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
@@ -114,36 +114,43 @@ abstract class AbstractProxy extends AbstractElementCollection implements IProxy
 	}
 
 	/**
-	 * Proxies get access of properties of the sub-object.
+	 * Proxies read access for properties of the sub-object.
 	 *
 	 * @param string $property Name of the property.
 	 *
 	 * @return mixed
-	 * @throws ElementException When sub-object doesn't have a specific property.
 	 */
 	public function __get($property)
 	{
-		$sub_object = $this->getObject();
+		$this->assertPropertyExistence($property);
 
-		if ( !property_exists($sub_object, $property) ) {
-			$message = sprintf('"%s" property is not available on the %s', $property, get_class($sub_object));
-
-			throw new ElementException($message, ElementException::TYPE_UNKNOWN_PROPERTY);
-		}
-
-		return $sub_object->{$property};
+		return $this->getObject()->$property;
 	}
 
 	/**
-	 * Proxies set access of properties of the sub-object.
+	 * Proxies write access for properties of the sub-object.
 	 *
 	 * @param string $property Name of the property.
 	 * @param mixed  $value    Value of the property.
 	 *
-	 * @return mixed
-	 * @throws ElementException When sub-object doesn't have a specific property.
+	 * @return void
 	 */
 	public function __set($property, $value)
+	{
+		$this->assertPropertyExistence($property);
+
+		$this->getObject()->$property = $value;
+	}
+
+	/**
+	 * Asserts existence of a property.
+	 *
+	 * @param string $property Name of the property.
+	 *
+	 * @return void
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	protected function assertPropertyExistence($property)
 	{
 		$sub_object = $this->getObject();
 
@@ -152,8 +159,6 @@ abstract class AbstractProxy extends AbstractElementCollection implements IProxy
 
 			throw new ElementException($message, ElementException::TYPE_UNKNOWN_PROPERTY);
 		}
-
-		$sub_object->{$property} = $value;
 	}
 
 	/**
--- a/library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
+++ b/library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
@@ -114,6 +114,49 @@ abstract class AbstractProxy extends AbstractElementCollection implements IProxy
 	}
 
 	/**
+	 * Proxies get access of properties of the sub-object.
+	 *
+	 * @param string $property Name of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __get($property)
+	{
+		$sub_object = $this->getObject();
+
+		if ( !property_exists($sub_object, $property) ) {
+			$message = sprintf('"%s" property is not available on the %s', $property, get_class($sub_object));
+
+			throw new ElementException($message, ElementException::TYPE_UNKNOWN_PROPERTY);
+		}
+
+		return $sub_object->{$property};
+	}
+
+	/**
+	 * Proxies set access of properties of the sub-object.
+	 *
+	 * @param string $property Name of the property.
+	 * @param mixed  $value    Value of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __set($property, $value)
+	{
+		$sub_object = $this->getObject();
+
+		if ( !property_exists($sub_object, $property) ) {
+			$message = sprintf('"%s" property is not available on the %s', $property, get_class($sub_object));
+
+			throw new ElementException($message, ElementException::TYPE_UNKNOWN_PROPERTY);
+		}
+
+		$sub_object->{$property} = $value;
+	}
+
+	/**
 	 * Locates element using the locator.
 	 *
 	 * @return NodeElement
--- a/library/QATools/QATools/PageObject/Proxy/IProxy.php
+++ b/library/QATools/QATools/PageObject/Proxy/IProxy.php
@@ -50,4 +50,25 @@ interface IProxy
 	 */
 	public function __call($method, array $arguments);
 
+	/**
+	 * Proxies get access of properties of the sub-object.
+	 *
+	 * @param string $property Name of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __get($property);
+
+	/**
+	 * Proxies set access of properties of the sub-object.
+	 *
+	 * @param string $property Name of the property.
+	 * @param mixed  $value    Value of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __set($property, $value);
+
 }
--- a/library/QATools/QATools/PageObject/Proxy/IProxy.php
+++ b/library/QATools/QATools/PageObject/Proxy/IProxy.php
@@ -51,23 +51,21 @@ interface IProxy
 	public function __call($method, array $arguments);
 
 	/**
-	 * Proxies get access of properties of the sub-object.
+	 * Proxies read access for properties of the sub-object.
 	 *
 	 * @param string $property Name of the property.
 	 *
 	 * @return mixed
-	 * @throws ElementException When sub-object doesn't have a specific property.
 	 */
 	public function __get($property);
 
 	/**
-	 * Proxies set access of properties of the sub-object.
+	 * Proxies write access for properties of the sub-object.
 	 *
 	 * @param string $property Name of the property.
 	 * @param mixed  $value    Value of the property.
 	 *
-	 * @return mixed
-	 * @throws ElementException When sub-object doesn't have a specific property.
+	 * @return void
 	 */
 	public function __set($property, $value);
 
--- a/tests/QATools/QATools/Live/PageObject/Element/GroupedInputContainer.php
+++ b/tests/QATools/QATools/Live/PageObject/Element/GroupedInputContainer.php
@@ -18,6 +18,14 @@ class GroupedInputContainer extends AbstractElementContainer
 {
 
 	/**
+	 * First example text input as WebElement.
+	 *
+	 * @var WebElement
+	 * @find-by('name' => 'test1')
+	 */
+	public $textInput1;
+
+	/**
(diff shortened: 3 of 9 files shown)


library/QATools/QATools/PageObject/Proxy/IProxy.php
```diff
@@ -50,4 +50,25 @@ public function getObject();
 	 */
 	public function __call($method, array $arguments);
 
+	/**
+	 * Proxies get access of properties of the sub-object.
```
aik099: There is no such `get access`, but there is `read access`.

library/QATools/QATools/PageObject/Proxy/IProxy.php
```diff
@@ -50,4 +50,25 @@ public function getObject();
...
+	 *
+	 * @param string $property Name of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __get($property);
+
+	/**
+	 * Proxies set access of properties of the sub-object.
```
aik099: There is no such `set access`, but there is `write access`.

library/QATools/QATools/PageObject/Proxy/IProxy.php
```diff
@@ -50,4 +50,25 @@ public function getObject();
...
+	 *
+	 * @param string $property Name of the property.
+	 * @param mixed  $value    Value of the property.
+	 *
+	 * @return mixed
```
aik099: Does setting element properly really cause some return value to be emitted?

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
```diff
@@ -114,6 +114,49 @@ public function __call($method, array $arguments)
...
+	{
+		$sub_object = $this->getObject();
+
+		if ( !property_exists($sub_object, $property) ) {
+			$message = sprintf('"%s" property is not available on the %s', $property, get_class($sub_object));
+
+			throw new ElementException($message, ElementException::TYPE_UNKNOWN_PROPERTY);
+		}
+
+		$sub_object->{$property} = $value;
```
aik099: The `{` and `}` are not required in this case.

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
```diff
@@ -114,6 +114,49 @@ public function __call($method, array $arguments)
...
+	{
+		$sub_object = $this->getObject();
+
+		if ( !property_exists($sub_object, $property) ) {
+			$message = sprintf('"%s" property is not available on the %s', $property, get_class($sub_object));
+
+			throw new ElementException($message, ElementException::TYPE_UNKNOWN_PROPERTY);
+		}
+
+		return $sub_object->{$property};
```
aik099: The `{` and `}` are not required in this case.

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
```diff
@@ -114,6 +114,49 @@ public function __call($method, array $arguments)
...
+	 * @param string $property Name of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __get($property)
+	{
+		$sub_object = $this->getObject();
+
+		if ( !property_exists($sub_object, $property) ) {
```
aik099: This IF is duplicated in `__set` method as well. Maybe it can be extracted to separate method, like `assertProperty` or similar.

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
```diff
@@ -114,6 +114,49 @@ public function __call($method, array $arguments)
...
+		return $sub_object->{$property};
+	}
+
+	/**
+	 * Proxies set access of properties of the sub-object.
+	 *
+	 * @param string $property Name of the property.
+	 * @param mixed  $value    Value of the property.
+	 *
+	 * @return mixed
```
aik099: Does setting element properly really cause some return value to be emitted?

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
```diff
@@ -114,6 +114,49 @@ public function __call($method, array $arguments)
...
 
 	/**
+	 * Proxies get access of properties of the sub-object.
+	 *
+	 * @param string $property Name of the property.
+	 *
+	 * @return mixed
+	 * @throws ElementException When sub-object doesn't have a specific property.
+	 */
+	public function __get($property)
```
aik099: Are you certain, that all proxies sub-class this `AbstractProxy` class and there this is the only class needs to be changed?

aik099:
Today’s review completed.

library/QATools/QATools/PageObject/Proxy/IProxy.php
evangelion1204: That was the word I was searching for :smile:

library/QATools/QATools/PageObject/Proxy/IProxy.php
evangelion1204: Nope, I will correct that.

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
evangelion1204: I just checked where we implemented `__call` and just did it the same way.

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
evangelion1204: Done

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
evangelion1204: Done

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
evangelion1204: Done

library/QATools/QATools/PageObject/Proxy/AbstractProxy.php
evangelion1204: Done

library/QATools/QATools/PageObject/Proxy/IProxy.php
evangelion1204: Done

library/QATools/QATools/PageObject/Proxy/IProxy.php
evangelion1204: Done

tests/QATools/QATools/Live/PageObject/Element/GroupedInputContainer.php
```diff
@@ -0,0 +1,30 @@
...
+use QATools\QATools\PageObject\Element\AbstractElementContainer;
+use QATools\QATools\PageObject\Element\WebElement;
+
+class GroupedInputContainer extends AbstractElementContainer
+{
+
+	/**
+	 * Example text inputs with multiple find-by annotations as WebElement.
+	 *
+	 * @var WebElement[]
```
aik099: Was there any reason to have property that will find several elements VS property, that contain 1 element?

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
```diff
@@ -140,6 +140,16 @@ public function testMethodForwardingFailure()
 		$this->element->nonExistingMethod();
 	}
 
+	/**
+	 * @expectedException \QATools\QATools\PageObject\Exception\ElementException
+	 * @expectedExceptionCode \QATools\QATools\PageObject\Exception\ElementException::TYPE_UNKNOWN_PROPERTY
+	 * @expectedExceptionMessage "nonExistingProperty" property is not available on the
```
aik099: Please specify full exception message here.

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
```diff
@@ -140,6 +140,16 @@ public function testMethodForwardingFailure()
...
+	 */
+	public function testPropertyForwardingFailure()
```
aik099: We're missing test that checks for successful property reading.

aik099:
Today’s review completed.

aik099:
What if you try to access private/protected property through proxy?

tests/QATools/QATools/Live/PageObject/Element/GroupedInputContainer.php
evangelion1204: Not really I just took the grouped item that where available, I can change the test to just match a single item.

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
evangelion1204: I just placed it there cause it was easier to implement. We already had a container where I could easily add a property. The only usual case to access a public property is a `AbstractContainer`.

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
evangelion1204: Which is not easy possible, cause the last part is element class, like in I just followed the `__call` message type. I could only add a RegExp to match a FQCN.

evangelion1204:
> If built-in PHP error for accessing protected/private property is too ugly, then maybe we should throw our own instead.

You are right, this will not check visibility. Any idea how to manage it besides reflection or setting a custom error handler?

tests/QATools/QATools/Live/PageObject/Element/GroupedInputContainer.php
aik099: Please do, because I don't want for tests to make an impression, that property access only works for multi-element properties.

evangelion1204: Ok now I have both, a single found element and multiple `find-by` elements. Both cases covered.

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
aik099: Then I guess we need success/failure tests for both element and container, because somebody might want to access properties on the element and not on element through container.

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
aik099: If that class name is auto-generated Mockery class name, then of course keep existing code. I thought it just real class name here.

aik099:
> Depending on the proxied element and usage of __get it might be even worth to use only a custom handler, property_exists and reflection will not help.

I haven't understood a thing from this. Sorry.

> You are right, this will not check visibility. Any idea how to manage it besides reflection or setting a custom error handler?

For us not to get performance issues on multiple property access I would get all public properties on an object right away on 1st property access attempt (the reflection can do scope filter automatically) and then just do `in_array` on each property access.

aik099:
@evangelion1204 , any update on this?

evangelion1204:
> I haven't understood a thing from this. Sorry.

Took me a while to figure out my comment ;) Reflection will help with normal properties and fetching them the first time via reflection and storing them is the best way. But whats about properties which are defined via the magic method `__get` like.

```php

class DynamicProperty
{
  public function __get($name)
  {
    switch($name) {
      case 'dynamic1':
        return ...
      case 'dynamic2':
        return ...
    }
  }
...
}
```

This case would not be coverable by reflection, only by custom error handler.

aik099:
I propose such approach:

1. use reflection to get all public properties
2. in case of `__get` or `__set` usage we look if reflected property exists and then return it
3. if reflected property doesn't exist, but `__get` or `__set` exists in proxied class, then we just call it and return it's result

Side note:
We might also use same approach with `__call` method (but in separate task of course) to allow proxied object to handle calls to unknown methods (if somebody will extend `WebElement` and create `__call` method in it).

aik099:
Today’s review completed. You can now implement the suggestions from other days and from today.

aik099:
> Side note:
We might also use same approach with `__call` method (but in separate task of course) to allow proxied object to handle calls to unknown methods (if somebody will extend `WebElement` and create `__call` method in it). 

Implemented in #169 and #172 .

aik099:
> https://github.com/qa-tools/qa-tools/pull/138#issuecomment-107714643

Same problem for `__call` method currently: #178

aik099:
@evangelion1204 , I'm just wondering why do we care if user accesses non-public property through proxy? For me it's perfectly OK if user will get back same error as if he would be accessing non-public property right on `WebElement` directly.

tests/QATools/QATools/PageObject/Proxy/AbstractProxyTestCase.php
aik099: I've now realized (by looking at the code), that this is a copy-paste from `__call` method testing. Sure you can keep this as-is.

aik099:
Closing in favor of #179. Your changes were incorporated into that PR.

PR #161, opened 2016-05-21
CONTRIBUTING.md update

slde-gorillaman:
Added a section to install Selenium and a test webserver in order to run the project's PhpUnit tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,37 @@ Make sure that you don't break anything with your changes by running:
 $> phpunit
 ```
 
+### Running Selenium and Testwebserver
+
+In order to run the Seleniumtests, a webserver and selenium have to be installed.
+You can do this by either running on your machine:
+
+```bash
+$> ./tests/install_webserver.sh
+$> ./tests/run_selenium.sh
+```
+
+or if you have docker installed:
+
+```bash
+$> docker run --name selenium-firefox -p 4444:4444 selenium/standalone-firefox
+$> docker run --name qa-tools-tests -p 80:80 -v ${PWD}:/usr/share/nginx/html:ro -d nginx
+```
+
+[More on Selenium Docker Image](https://github.com/SeleniumHQ/docker-selenium)
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -26,6 +26,37 @@ Make sure that you don't break anything with your changes by running:
...
+
+```bash
+$> ./tests/install_webserver.sh
+$> ./tests/run_selenium.sh
+```
+
+or if you have docker installed:
+
+```bash
+$> docker run --name selenium-firefox -p 4444:4444 selenium/standalone-firefox
```
aik099: Actually this way (via docker) of installing Selenium should be mentioned on https://github.com/qa-tools/phpunit-example/blob/master/README.md page.

CONTRIBUTING.md
```diff
@@ -26,6 +26,37 @@ Make sure that you don't break anything with your changes by running:
 $> phpunit
 ```
 
+### Running Selenium and Testwebserver
+
+In order to run the Seleniumtests, a webserver and selenium have to be installed.
+You can do this by either running on your machine:
+
+```bash
+$> ./tests/install_webserver.sh
```
aik099: These files are used during build and

* download selenium server each time 
* start new web server

In real life however developers would have web server running anyway.

To summarize I recommend to:

1. go to https://github.com/qa-tools/phpunit-example
2. improve selenium installation instructions by
 * adding case when Docker is used
 * adding instruction on creating `phpunit.xml` file with right env variables
3. copy paste relevant instructions in this PR

aik099:
I guess you just wanted to rename the branch, but that lead GitHub think that PR needs to be closed.

slde-gorillaman:
Na, I closed it because I didn't see a need for the added documentation after all :)

aik099:
Actually I thought, that you'll do something like this: b5189dd0d346ffe6fd518b81843e82fa8cbed2ea (that is what I did now), because instructions for configuring/running Selenium Server were really missing on the contributions page.

PR #200, opened 2017-10-01
Upgrade Selenium from 2.5.1 to 3.6.0

aik099:


peter279k:
This CI build failed because the web driver is failed to start in doing unit tests.

It should check that error deeply.

aik099:
I've totally forgot about this PR myself (almost 2 years now). If you'd like you can continue or send alternative PR.

I believe the library itself works with Selenium 3 (because I'm using only what Mink allows), but tests are needed just to be certain.

aik099:
Closing in favor of #205

Issue #196, opened 2016-10-07
[Insight] Boolean should be compared strictly - in DefaultPropertyDecorator.php:83


aik099:
**[in library/QATools/QATools/PageObject/PropertyDecorator/DefaultPropertyDecorator.php, line 83](https://github.com/qa-tools/qa-tools/blob/505b14a7766daa718cad21bdb7bcb404cabaa1fc/library/QATools/QATools/PageObject/PropertyDecorator/DefaultPropertyDecorator.php#L83)**

> With booleans, only strict comparison (with `!==` operator) should be used to lower bug risks and to improve performances.

``` php
            return null;
        }

        $locator = $this->locatorFactory->createLocator($property);

        return $locator != null ? $this->doDecorate($property, $locator) : null;
    }

    /**
     * Checks if a property can be decorated.
     *
```

---

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/98070f4a-70bd-49a0-81b2-7c47caac0274/analyses/5)_

michaeltintiuc:
@aik099 Do you mean comparison to `null` what do you think of `is_null` function? Would be happy to help out

aik099:
That's not my words, but a bot who posted the issue. Yes, it wants `!==` instead of `!=`.

michaeltintiuc:
Sounds great, would ti be possible to assign me to this issue and I'd tackle it over the weekend

aik099:
Sure, take it. Unfortunately I can't specify you in "Assignee" field of the issue, because GitHub only shows users with PUSH access there.

michaeltintiuc:
Thanks @aik099 will send a PR over the weekend!

aik099:
You can also look at other tasks in this repo if you like. I can mark the ones you require with Hacktoberfest label if needed.

michaeltintiuc:
Yeah, would be great I skimmed through some

On Sep 28, 2017 19:01, "Alexander Obuhovich" <notifications@github.com>
wrote:

You can also look at other tasks in this repo if you like. I can mark the
ones you require with Hacktoberfest label if needed.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/qa-tools/qa-tools/issues/196#issuecomment-332883486>,
or mute the thread

aik099:
Fixed in #198.

PR #201, opened 2017-10-01
Replace all is_null occurances with strict comparison

michaeltintiuc:
Changes proposed as per discussion in issue #196


aik099:
It appears, that there is a communication issue in here. In the #196 you've replaced `!= null` with `!== null` and told that there are more places like this.

Replacing `is_null` with `!== null` isn't something that is needed at all.

michaeltintiuc:
Ah, I see, no problem, perhaps it made more sense to use `is_null` in the previous PR then

aik099:
Who knows. I think we should keep `is_null` and `!== null` check in their places without changing for now. Thank you.

